PR #9, opened 2016-08-26
Search fail

### LINKED ISSUE #8: Search might fail
The "name search" in the stops index has an auto-complete on it, so it's less likely that this will happen, but if the user enters a name that doesn't match exactly the name of an existing stop they'll get a 500 error [here] (because `stop` will be `nil`).

There should be an error behavior too.

[1]:  https://github.com/umts/stop-project/blob/master/app/controllers/bus_stops_controller.rb#L36

mboneil10:
Closes #8

### DIFF
--- a/app/controllers/bus_stops_controller.rb
+++ b/app/controllers/bus_stops_controller.rb
@@ -33,7 +33,12 @@ class BusStopsController < ApplicationController
 
   def name_search
     stop = BusStop.find_by name: params.require(:name)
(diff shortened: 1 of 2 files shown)


mboneil10:
The page can't be viewed because of an issue with the haml. I'll fix that while I'm here.

mboneil10:
screen shot 2016-08-26 at 4 42 06 pm

dfaulken:
Good work.

PR #11, opened 2017-01-05
Add timestamps to BusStop

### LINKED ISSUE #10: Add BusStop timestamps
@sherson suggested that default timestamps would be useful here. I didn't think they were needed originally.

dfaulken:
Also add completion timestamp, and mark as completed when completed is checked in form.

Closes #10.

### DIFF
--- a/db/migrate/20170105165853_add_timestamps_to_bus_stops.rb
+++ b/db/migrate/20170105165853_add_timestamps_to_bus_stops.rb
@@ -1,0 +1,9 @@
+class AddTimestampsToBusStops < ActiveRecord::Migration
+  def change
+    change_table :bus_stops do |t|
+      t.timestamps null: false
+      t.boolean :completed
+      t.datetime :completed_at
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,33 +11,37 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160708192840) do
+ActiveRecord::Schema.define(version: 20170105165853) do
 
   create_table "bus_stops", force: :cascade do |t|
-    t.string  "name",                      limit: 255, default: "", null: false
-    t.string  "hastus_id",                 limit: 255,              null: false
-    t.string  "accessible",                limit: 255
-    t.string  "bench",                     limit: 255
-    t.string  "curb_cut",                  limit: 255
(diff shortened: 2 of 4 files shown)


db/schema.rb
```diff
@@ -11,33 +11,37 @@
...
+    t.boolean  "bus_pull_out_exists"
+    t.boolean  "extend_pole"
+    t.boolean  "has_power"
+    t.boolean  "new_anchor"
+    t.boolean  "new_pole"
+    t.boolean  "solar_lighting"
+    t.boolean  "straighten_pole"
+    t.boolean  "system_map_exists"
+    t.integer  "mounting_clearance_after",  limit: 4
+    t.integer  "mounting_clearance_before", limit: 4
```
dfaulken: Everything upwards of here is not a change. The `limit` values were automatically pushed over by one character (apparently).

db/migrate/20170105165853_add_timestamps_to_bus_stops.rb
```diff
@@ -0,0 +1,9 @@
+class AddTimestampsToBusStops < ActiveRecord::Migration
+  def change
+    change_table :bus_stops do |t|
+      t.timestamps null: false
```
werebus: Does this cause any problems with existing data?  That is, does AR fill in the `created_at` with... something if its missing?

db/schema.rb
```diff
@@ -65,6 +69,7 @@
     t.string   "current_sign_in_ip",     limit: 255
     t.string   "last_sign_in_ip",        limit: 255
     t.boolean  "admin",                              default: false, null: false
+    t.datetime "created_at"
```
werebus: I wonder where this came from

db/migrate/20170105165853_add_timestamps_to_bus_stops.rb
dfaulken: Yes, apparently they get filled in with the time at which the migration is run.

db/schema.rb
dfaulken: Oh weird, I didn't notice that. I was doing some migrating and schema loading to resolve some issues with my development database, so that was probably the cause.

db/migrate/20170105165853_add_timestamps_to_bus_stops.rb
werebus: That makes sense to me.

PR #24, opened 2018-03-20
Manage stops

### LINKED ISSUE #18: Manage stops
Would be useful to know when stops have been updated. Some common uses:

* How many stops have not been updated since X date
* A list of the stops that have not been updated since X date

Might be more flexible to provide:

* How many stops were updated over a date range
* A list of stops that were updated over a date range

Or a CSV export

dfaulken:
Closes #18.

Link to new interface in navbar:

screen shot 2018-03-20 at 5 41 58 pm

Management interface:

screen shot 2018-03-20 at 5 42 16 pm

Includes pagination:

screen shot 2018-03-20 at 5 42 31 pm

Can view outdated stops:

screen shot 2018-03-20 at 5 42 48 pm

Including a datepicker:

screen shot 2018-03-20 at 5 43 18 pm

And can export as CSV:

screen shot 2018-03-20 at 5 43 43 pm

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ gem 'paper_trail', '~> 8.1'
 gem 'rails', '~> 5.1'
 gem 'sass-rails', '~> 5.0'
 gem 'underscore-rails', '~> 1.8'
+gem 'will_paginate', '~> 3.1'
 
 group :production do
   gem 'exception_notification', '~> 4.2'
--- a/Gemfile
+++ b/Gemfile
@@ -24,4 +24,5 @@ group :development do
   gem 'capistrano-rails', require: false
   gem 'capistrano-passenger', require: false
   gem 'pry-byebug'
+  gem 'timecop', '~> 0.9.1'
 end
--- a/app/assets/javascripts/bus_stops.js
+++ b/app/assets/javascripts/bus_stops.js
@@ -2,6 +2,10 @@ $(document).ready(function(){
   $('.name-autocomplete').autocomplete({
     source: searchStops
   });
+
+  $('.datepick').datepicker({
+    dateFormat: 'yyyy-mm-dd'
+  })
 });
 
 function searchStops(request, response){
--- a/app/assets/javascripts/bus_stops.js
+++ b/app/assets/javascripts/bus_stops.js
@@ -4,7 +4,7 @@ $(document).ready(function(){
   });
 
   $('.datepick').datepicker({
-    dateFormat: 'yyyy-mm-dd'
+    dateFormat: 'yy-mm-dd'
   })
 });
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,5 +1,5 @@
 /*
- *= require autocomplete
+ *= require jquery-ui
  *= require_tree .
  *= require_self
  */
--- a/app/controllers/bus_stops_controller.rb
+++ b/app/controllers/bus_stops_controller.rb
@@ -1,5 +1,6 @@
 class BusStopsController < ApplicationController
-  before_action :find_stop, only: %i(edit id_search update)
+  before_action :find_stop, only: %i[edit id_search update]
+  before_action :restrict_to_admin, only: %i[manage]
 
   def autocomplete
     stops = BusStop.where 'lower(name) like ?',
@@ -29,6 +30,10 @@ class BusStopsController < ApplicationController
 
   def id_search
     redirect_to edit_bus_stop_path(@stop.hastus_id)
+  end
+
+  def manage
+    @stops = BusStop.order(:name)
   end
 
   def name_search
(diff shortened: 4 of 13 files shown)


dfaulken:
Can update all as generic CSV:

screen shot 2018-03-20 at 5 53 54 pm

Link:

screen shot 2018-03-20 at 5 54 10 pm

Gemfile
```diff
@@ -23,4 +24,5 @@ group :development do
   gem 'capistrano-rails', require: false
   gem 'capistrano-passenger', require: false
   gem 'pry-byebug'
+  gem 'timecop', '~> 0.9.1'
```
dfaulken: Timecop is used in seeding to simulate incidents having been created/updated at different times in the past.

app/assets/javascripts/bus_stops.js
```diff
@@ -2,6 +2,10 @@ $(document).ready(function(){
   $('.name-autocomplete').autocomplete({
     source: searchStops
   });
+
+  $('.datepick').datepicker({
+    dateFormat: 'yy-mm-dd'
```
dfaulken: It *looks* like a two-digit year, but it's actually four. If you write `yyyy` you end up with `20182018`.

app/assets/stylesheets/application.scss
```diff
@@ -1,5 +1,5 @@
 /*
- *= require autocomplete
+ *= require jquery-ui
```
dfaulken: Just all of it, why not

app/controllers/bus_stops_controller.rb
```diff
@@ -1,16 +1,17 @@
 class BusStopsController < ApplicationController
-  before_action :find_stop, only: %i(edit id_search update)
+  before_action :find_stop, only: %i[edit destroy id_search update]
+  before_action :restrict_to_admin, only: %i[destroy manage]
 
   def autocomplete
     stops = BusStop.where 'lower(name) like ?',
                           "%#{params.require(:term)}%"
-    render json: stops.pluck(:name)
+    render json: stops.pluck(:name).sort
```
dfaulken: Due to removing default scopes (I don't like them anymore)

app/models/bus_stop.rb
```diff
@@ -50,6 +60,23 @@ def last_updated_by
...
   end
 
+  def route_list
+    routes.pluck(:number).sort.join(', ')
+  end
+
+  def self.to_csv(limited_attributes: false)
+    attrs = if limited_attributes
+              LIMITED_ATTRS
+            else Hash[columns.map{|c| [c.name, c.name.humanize] }]
```
dfaulken: Should I use `human_attribute_name` here to respect locales in the future?

app/models/bus_stop.rb
```diff
@@ -50,6 +60,23 @@ def last_updated_by
...
+            end
+    CSV.generate headers: true do |csv|
+       csv << attrs.values
+       all.each do |stop|
+         csv << attrs.keys.map { |attr| stop.send attr }
+       end
```
dfaulken: I wanted to do

```ruby
csv += all.map do |stop|
  attrs.keys.map { |attr| stop.send.attr }
end
```,

but `CSV` doesn't implement `+=`.

File: app/views/bus_stops/outdated.haml [hunk 7]
```diff
@@ -0,0 +1,26 @@
+Bus stops not updated since #{@date} (#{@stops.count})
+
+= form_tag '#', method: :get do
+  = text_field_tag :date, @date, class: 'datepick'
+  = submit_tag 'Change date'
+
+= link_to 'Export as CSV', outdated_bus_stops_path(date: params[:date], format: :csv)
```
dfaulken: I tried `..._path(params, format: :csv)` for future proofing sake, but that didn't work.

app/models/bus_stop.rb
Anbranin: I wouldn't worry about it.

dfaulken:
@sherson does the interface look good to you?

app/assets/javascripts/bus_stops.js
mboneil10: Waaaaaat

app/controllers/bus_stops_controller.rb
```diff
@@ -40,6 +59,20 @@ def name_search
     end
   end
 
+  def outdated
+    @date = Date.parse(params[:date]) rescue 1.month.ago.to_date
```
mboneil10: Why is the rescued value `1.month.ago.to_date`?

app/models/bus_stop.rb
mboneil10: :(

app/controllers/bus_stops_controller.rb
dfaulken: It needs to be something. I'm open to suggestions.

dfaulken:
> Do we need to use console magic to restore a mistakenly deleted stop?

Yes. Stops are paper-trailed, so that would be easiest.

PR #35, opened 2018-05-25
All incomplete

### LINKED ISSUE #28: Mark every Stop as Incomplete
Requested during PVTA meeting on April 30th, 2018:

Make a way to mark every stop as incomplete. @dfaulken and I decided that this should be a rake task.

mboneil10:
Closes #28


mboneil10:
@dfaulken namespaces for the sake of organization/clarity, right? Is there another reason why we use namespaces?

lib/tasks/all_incomplete.rake
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+namespace :update_all_stops do
+  desc "Mark all stops as incomplete"
+  task :all_incomplete do
```
werebus: Probably need `:environment` here, right?

mboneil10: Good call. We're using `uncomplete_all`. Unless anyone has any other ideas.

mboneil10:
I changed the filename although no one had a problem with it... For the sake of future programmers/rake tasks.

PR #42, opened 2018-06-01
Display name

### LINKED ISSUE #36: Using Display_Name for More than Stops
Why does [this line work](https://github.com/umts/stop-project/blob/d069029d4805774f603d676a841b286760f82298/app/views/users/edit.haml#L3) if the method `display_name` is [expecting a stop](https://github.com/umts/stop-project/blob/d069029d4805774f603d676a841b286760f82298/app/helpers/bus_stops_helper.rb#L16)?

mboneil10:
Closes #36

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,9 +1,9 @@
 module ApplicationHelper
 
   def display_name(item)
-    case item
-    when :user then item.name.present? ? item.name : 'New user'
-    when :stop then item.name.present? ? item.name : 'New stop'
+    case item.class
+    when User then item.name.present? ? item.name : 'New user'
+    when BusStop then item.name.present? ? item.name : 'New stop'
     end
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,9 +1,11 @@
 module ApplicationHelper
 
   def display_name(item)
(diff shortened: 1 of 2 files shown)


mboneil10:
Tried to do the `item_type` thing? Couldn't figure it out. Please help.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,9 @@
 module ApplicationHelper
+
+  def display_name(item)
+    case item.class
```
dfaulken: Case statements evaluate with `===`, so you can do things like

```ruby
case some_variable
when String
  puts some_variable
when Integer
  some_variable.succ.to_s
#...
end
```

dfaulken: Without needing to say `some_variable.class`, that is.

mboneil10: Thanks, I think I've only ever written one once before. Not a real excuse, but thanks!! I'll make the change now.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,11 @@
 module ApplicationHelper
+
+  def display_name(item)
+    case item
+    when User
+      item.name.present? ? item.name : 'New user'
```
dfaulken: How about `item.name || 'New user'`?

PR #43, opened 2018-06-01
Change HASTUS fields and selects into text

### LINKED ISSUE #41: User can change stop Name and Hastus ID
We should probably prevent this.

mboneil10:
Closes #41


mboneil10:
Makes it impossible to add `name`, `hastus_id`, and `route_list` in a new stop. But it was decided in the PVTA meeting yesterday that we didn't want anyone to be able to create stops on this website. Stops should be created in Hastus.

jforgue:
I like how you're displaying the routes now, it doesn't take up too much space.

Issue #21, opened 2018-03-17
Tests


dfaulken:
Didn't write tests for this. Wasn't necessary for MVP in the agile cycle.

Should backfill them continually.

dfaulken:
Or, could wait until initial set of meetings about this project codifies the desired functionality to the point where writing tests won't be a waste.

mboneil10:
I'll get started on integration tests.

mboneil10:
Here's the plan, 3 different PRS:
- test setup
- unit tests
- system tests

mboneil10:
We're pretty good with tests. I expect coverage to bump once we go deeper into the rewrite.

PR #47, opened 2018-06-05
System tests

mboneil10:
Part 2 of 3 for #21

### DIFF
--- a/config/initializers/paper_trail.rb
+++ b/config/initializers/paper_trail.rb
@@ -1,0 +1,1 @@
+PaperTrail.config.track_associations = false
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 45.83
+    "covered_percent": 19.91
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 19.91
+    "covered_percent": 21.72
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 23.08
+    "covered_percent": 22.62
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 67.27
+    "covered_percent": 21.79
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 48.21
+    "covered_percent": 23.08
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 21.79
+    "covered_percent": 22.62
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 21.72
+    "covered_percent": 83.44
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 22.62
+    "covered_percent": 48.21
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 22.62
+    "covered_percent": 23.08
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 38.69
+    "covered_percent": 45.83
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 23.08
+    "covered_percent": 38.69
   }
 }
--- a/spec/factories/bus_stops.rb
+++ b/spec/factories/bus_stops.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :bus_stop do
     name 'stop'
-    sequence(:hastus_id)
+    sequence :hastus_id
   end
 end
--- a/spec/factories/bus_stops.rb
+++ b/spec/factories/bus_stops.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 FactoryBot.define do
   factory :bus_stop do
     name 'stop'
--- a/spec/system/bus_stops/managing_bus_stops_spec.rb
+++ b/spec/system/bus_stops/managing_bus_stops_spec.rb
@@ -62,9 +62,3 @@ describe 'viewing outdated' do
     pending('review datepickers and capybara')
   end
 end
-
-describe 'exporting all as CSV' do
-  it 'exports a CSV of all stops' do
-    pending('maybe later')
-  end
-end
--- a/spec/system/bus_stops/managing_bus_stops_spec.rb
+++ b/spec/system/bus_stops/managing_bus_stops_spec.rb
@@ -4,7 +4,7 @@ describe 'managing stops as an admin' do
 
   before :each do
     admin = create :user, :admin
-    @bus_stop = create :bus_stop, name: 'GRC'
+    @bus_stop = create :bus_stop
     
     when_current_user_is admin
     visit manage_bus_stops_url
--- a/spec/system/bus_stops/managing_bus_stops_spec.rb
+++ b/spec/system/bus_stops/managing_bus_stops_spec.rb
@@ -1,47 +1,45 @@
 require 'spec_helper'
 
 describe 'managing stops as an admin' do
+  let(:admin) { create :user, :admin }
+  let!(:bus_stop) { create :bus_stop }
   before :each do
-    admin = create :user, :admin
-    @bus_stop = create :bus_stop
-    
     when_current_user_is admin
     visit manage_bus_stops_url
   end
   context 'delete button' do
     it 'deletes the specific bus stop' do
       expect(page).to have_selector 'table.manage tbody tr', count: 1
-      within 'tr', text: @bus_stop.name do
+      within 'tr', text: bus_stop.name do
         click_button 'Delete'
       end
       page.driver.browser.switch_to.alert.accept
       expect(page).to have_selector 'p.notice',
-        text: "#{@bus_stop.name} has been deleted."
-      expect(page).not_to have_selector 'table.manage tbody tr', text: @bus_stop.name
+        text: "#{bus_stop.name} has been deleted."
+      expect(page).not_to have_selector 'table.manage tbody tr', text: bus_stop.name
     end
   end
   context 'edit button' do
     it 'redirects to edit bus stop page' do
       expect(page).to have_selector 'table.manage tbody tr', count: 1
-      within 'tr', text: @bus_stop.name do
+      within 'tr', text: bus_stop.name do
         click_link 'Edit'
       end
-      expect(page).to have_content "Editing #{@bus_stop.name}"
-      expect(page.current_url).to end_with edit_bus_stop_path(@bus_stop.hastus_id)
+      expect(page).to have_content "Editing #{bus_stop.name}"
+      expect(page.current_url).to end_with edit_bus_stop_path(bus_stop.hastus_id)
     end
   end
 end
 
 describe 'viewing outdated' do
+  let(:admin) { create :user, :admin }
+  let!(:present_stop) { create :bus_stop }
+  let!(:date) { Date.today }
+  let!(:old_stop_1) { create :bus_stop, updated_at: (date - 2.months) }
+  let!(:old_stop_2) { create :bus_stop, updated_at: (date - 3.months) }
   before :each do
-    admin = create :user, :admin
-    @present_stop = create :bus_stop
-    @date = Date.today
-    @stop_1 = create :bus_stop, updated_at: (@date - 2.months)
-    @stop_2 = create :bus_stop, updated_at: (@date - 3.months)
     when_current_user_is admin
     visit manage_bus_stops_url
-
     click_link 'View Outdated'
   end
   it 'redirects to outdated page' do
@@ -49,14 +47,14 @@ describe 'viewing outdated' do
   end
   it 'displays only outdated stops' do
     expect(page).to have_selector 'table.manage tbody tr', count: 2
-    expect(page).to have_selector 'table.manage tbody tr', text: @stop_1.name
-    expect(page).to have_selector 'table.manage tbody tr', text: @stop_2.name
+    expect(page).to have_selector 'table.manage tbody tr', text: old_stop_1.name
+    expect(page).to have_selector 'table.manage tbody tr', text: old_stop_2.name
   end
   it 'allows editing of outdated stops' do
-    within 'tr', text: @stop_1.hastus_id do
+    within 'tr', text: old_stop_1.hastus_id do
       click_link 'Edit'
     end
-    expect(page).to have_content "Editing #{@stop_1.name}"
+    expect(page).to have_content "Editing #{old_stop_1.name}"
   end
   it 'outdated can be narrowed down with a different date' do
     pending('review datepickers and capybara')
--- a/spec/system/bus_stops/managing_bus_stops_spec.rb
+++ b/spec/system/bus_stops/managing_bus_stops_spec.rb
@@ -59,16 +59,12 @@ describe 'viewing outdated' do
     expect(page).to have_content "Editing #{@stop_1.name}"
   end
   it 'outdated can be narrowed down with a different date' do
-    within '#date' do
-      # datepickers guh
-      fill_in 'date', with: @date
-      binding.pry
-      click_button 'Change date'
-    end
+    pending('review datepickers and capybara')
   end
 end
 
 describe 'exporting all as CSV' do
   it 'exports a CSV of all stops' do
+    pending('maybe later')
   end
 end
--- a/spec/system/bus_stops/managing_bus_stops_spec.rb
+++ b/spec/system/bus_stops/managing_bus_stops_spec.rb
@@ -47,12 +47,12 @@ describe 'viewing outdated' do
   end
   it 'displays only outdated stops' do
     expect(page).to have_selector 'table.manage tbody tr', count: 2
-    expect(page).to have_selector 'table.manage tbody tr', text: old_stop_1.updated_at
-    expect(page).to have_selector 'table.manage tbody tr', text: old_stop_2.updated_at
-    expect(page).not_to have_selector 'table.manage tbody tr', text: present_stop.updated_at
+    expect(page).to have_selector 'table.manage tbody tr', text: old_stop_1.updated_at.strftime('%Y-%m-%d %H:%M %P')
+    expect(page).to have_selector 'table.manage tbody tr', text: old_stop_2.updated_at.strftime('%Y-%m-%d %H:%M %P')
+    expect(page).not_to have_selector 'table.manage tbody tr', text: present_stop.updated_at.strftime('%Y-%m-%d %H:%M %P')
   end
   it 'allows editing of outdated stops' do
-    within 'tr', text: old_stop_1.updated_at do
+    within 'tr', text: old_stop_1.updated_at.strftime('%Y-%m-%d %H:%M %P') do
       click_link 'Edit'
     end
     expect(page).to have_content "Editing #{old_stop_1.name}"
--- a/spec/system/bus_stops/managing_bus_stops_spec.rb
+++ b/spec/system/bus_stops/managing_bus_stops_spec.rb
@@ -33,15 +33,32 @@ describe 'managing stops as an admin' do
 end
 
 describe 'viewing outdated' do
-  context 'click the link' do
-    it 'redirects to outdated page' do
-      click_link 'View Outdated'
-      expect(page.current_url).to end_with outdated_bus_stops_path
+  before :each do
+    admin = create :user, :admin
+    @present_stop = create :bus_stop
+    @date = Date.today
+    @stop_1 = create :bus_stop, updated_at: (@date - 2.months)
+    @stop_2 = create :bus_stop, updated_at: (@date - 3.months)
+    when_current_user_is admin
+    visit manage_bus_stops_url
+
+    click_link 'View Outdated'
+  end
+  it 'redirects to outdated page' do
+    expect(page.current_url).to end_with outdated_bus_stops_path
+  end
+  it 'displays only outdated stops' do
+    expect(page).to have_selector 'table.manage tbody tr', count: 2
+    expect(page).to have_selector 'table.manage tbody tr', text: @stop_1.name
+    expect(page).to have_selector 'table.manage tbody tr', text: @stop_2.name
+  end
+  it 'allows editing of outdated stops' do
+    within 'tr', text: @stop_1.hastus_id do
+      click_link 'Edit'
     end
-    it 'allows editing of outdated stops' do
-    end
-    it 'outdated can be narrowed down with a different date' do
-    end
+    expect(page).to have_content "Editing #{@stop_1.name}"
+  end
+  it 'outdated can be narrowed down with a different date' do
   end
 end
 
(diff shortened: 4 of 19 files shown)


mboneil10:
Hi all reviewers, I'm aware this PR is long and awful. I thought that since this a small app, there wouldn't be too many system tests. I was wrong. And, I haven't even finished. Some index tests are failing, I've left three tests pending, and I didn't even touch the edit stops feature.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 5.6
+    "covered_percent": 38.69
```
mboneil10: But this is what it says after I run rspec: `129 / 155 LOC (83.23%) covered.`??

spec/factories/bus_stops.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :bus_stop do
+    name 'stop'
```
dfaulken: Stops can have the same name?

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    when_current_user_is admin
+    visit manage_bus_stops_url
+  end
+  context 'delete button' do
+    it 'deletes the specific bus stop' do
+      expect(page).to have_selector 'table.manage tbody tr', count: 1
+      within 'tr', text: @bus_stop.name do
+        click_button 'Delete'
+      end
+      page.driver.browser.switch_to.alert.accept
```
dfaulken: Could you walk through what this does? Haven't seen this before in our system tests.

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    expect(page).to have_content "Editing #{@stop_1.name}"
+  end
+  it 'outdated can be narrowed down with a different date' do
+    pending('review datepickers and capybara')
+  end
+end
+
+describe 'exporting all as CSV' do
+  it 'exports a CSV of all stops' do
+    pending('maybe later')
```
dfaulken: I think it's fine for these to be unit tests.

spec/system/index_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      expect(page.current_url).to end_with bus_stops_path
+    end
+    it 'displays a helpful message' do
+      expect(page).to have_selector 'p.notice',
+        text: 'Stop 1234 not found'
+    end
+  end
+  context 'without completing stop id' do
+    it 'autofills' do
+      pending('how do I check this')
```
dfaulken: I might start by pausing execution at the point where you expect an autocomplete to be working. Then examine the DOM and see what you can work with.

spec/system/index_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+  end
+end
+
+describe 'searching for a bus stop by route' do
+  before :each do
+    user = create :user
+    @route = create :route
+    @bus_stop = create :bus_stop
+    @route.bus_stops << @bus_stop
```
dfaulken: You may be able to do this as `create :bus_stop, routes: [@route]`.

spec/system/index_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  end
+  context 'without completing stop id' do
+    it 'autofills' do
+      pending('how do I check this')
+    end
+  end
+end
+
+describe 'searching for a bus stop by route' do
+  before :each do
```
dfaulken: Any particular reason to use `before :each` and instance variables rather than `let` blocks?

spec/system/index_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    @route = create :route
+    @bus_stop = create :bus_stop
+    @route.bus_stops << @bus_stop
+    when_current_user_is user
+    visit root_url
+  end
+  context 'route from the dropdown' do
+    it 'redirects to bus stops by route page' do
+      within 'form', text: 'Select a route' do
+        select @route.number, from: 'number'
```
dfaulken: I think the argument to `from` ought to be one of two things:

1. The ID of the element, as a symbol (less good);
2. The label of the element, as a string (best).

In general, your system tests should really only be based around what people can see and do, rather than things like the DOM structure — as much as possible, that is.

mboneil10:
Also, the deprecation warning I'm talking about in 0d86eca95ed7639c2b6116f7ffc8d52e342590b8 is this:

> DEPRECATION WARNING: PaperTrail.config.track_associations has not been set. As of PaperTrail 5, it defaults to false. Tracking associations is an experimental feature so we recommend setting PaperTrail.config.track_associations = false in your config/initializers/paper_trail.rb  (called from block (2 levels) in <top (required)> at /Users/molly/stop-project/spec/system/index_spec.rb:41)

coverage/.last_run.json
mboneil10: @dfaulken idk why the `covered_percent` in `.last_run.json` is 38.69 but after running rspec, it shows up in the terminal as `83.28`

spec/factories/bus_stops.rb
mboneil10: [Yeah, at least that's what's written](https://github.com/umts/stop-project/blob/master/app/models/bus_stop.rb#L7)

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: It accepts the js alert box

spec/system/index_spec.rb
mboneil10: No, should I try to completely move over from instance variables to `let` blocks?

spec/factories/bus_stops.rb
mboneil10: I don't see how that would be useful though

spec/system/index_spec.rb
mboneil10: OMG it work. Wow, rails. wow.

spec/factories/bus_stops.rb
dfaulken: @sherson, any thoughts about whether we can expect stops to have unique names coming from Hastus?

spec/system/bus_stops/managing_bus_stops_spec.rb
dfaulken: Ah, okay. I suppose I could have guessed that.

If we think we might reuse this, it could be worth putting this in a helper method. That would also give us an opportunity to ascribe a descriptive name to the code.

It's a little unusual to see six short methods chained like that, is why it caught my attention.

spec/system/index_spec.rb
dfaulken: Well, it's been the convention we've moved to in recent times, but that doesn't necessarily mean that it's the best fit for every task.

What are your thoughts?

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: Alright, sounds good

spec/system/index_spec.rb
mboneil10: My thoughts are silly about this.

mboneil10: Not a good programmer thing...

mboneil10: It's easier for me to spot instance variables because they're highlighted. Literally the only reason I use them here. I'll switch them over since there's no point in keeping it like this.

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    when_current_user_is admin
+    visit manage_bus_stops_url
+    click_link 'View Outdated'
+  end
+  it 'redirects to outdated page' do
+    expect(page.current_url).to end_with outdated_bus_stops_path
+  end
+  it 'displays only outdated stops' do
+    expect(page).to have_selector 'table.manage tbody tr', count: 2
+    expect(page).to have_selector 'table.manage tbody tr', text: old_stop_1.updated_at.strftime('%Y-%m-%d %H:%M %P')
```
mboneil10: Any thoughts on how to make this less disgusting? I was thinking an instance variable.

mboneil10:
Still need to fill in date picker and autofill for search test.

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:admin) { create :user, :admin }
+  let!(:bus_stop) { create :bus_stop }
+  before :each do
+    when_current_user_is admin
+    visit manage_bus_stops_url
+  end
+  context 'delete button' do
+    it 'deletes the specific bus stop' do
+      expect(page).to have_selector 'table.manage tbody tr', count: 1
+      within 'tr', text: bus_stop.name do
```
jforgue: This is a cool way to do this, I haven't seen a `within` block in a test before. You mentioned that `bus_stop` names don't need to be unique though. Actually you set them to be the same at creation. So, this wouldn't actually help specify what button to press.

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        click_button 'Delete'
+      end
+      page.driver.browser.switch_to.alert.accept
```
jforgue: You could set `page.driver.browser.switch_to.alert` to a variable to make it's value more clear.

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(page).to have_selector 'p.notice',
+        text: "#{bus_stop.name} has been deleted."
+      expect(page).not_to have_selector 'table.manage tbody tr', text: bus_stop.name
```
jforgue: Again the name may not be unique. I think you get my point, I'll stop commenting on names.

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+end
+
+describe 'viewing outdated' do
+  let(:admin) { create :user, :admin }
+  let!(:present_stop) { create :bus_stop }
+  let!(:date) { Date.today }
+  let!(:old_stop_1) { create :bus_stop, updated_at: (date - 2.months) }
+  let!(:old_stop_2) { create :bus_stop, updated_at: (date - 3.months) }
+  before :each do
```
jforgue: If you're going to have a `before :each` I think you should define your variables in it rather than `let` blocks. I don't think there's any need for both in this case.

spec/system/bus_stops/managing_bus_stops_spec.rb
jforgue: You could write a helper method, but that seems a little overkill.

spec/system/index_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    when_current_user_is user
+    visit root_url
+  end
+  context 'correct stop id' do
+    it 'redirects to the edit page' do
+      within 'form', text: 'Enter stop ID' do
+        fill_in 'Enter stop ID', with: bus_stop.hastus_id.to_s
+        click_button 'Search'
+      end
+      expect(page).to have_content "Editing #{bus_stop.name}"
```
jforgue: I'm unfamiliar with this app so I'm curious. Does clicking search only have the options of bringing up the edit page for a specific stop or telling you the stop doesn't exist? Usually searches give you a list of possible options to select from. This is less of a "search" and more of a "find" I guess.

spec/system/index_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+     it 'stays on the same page' do
+       expect(page.current_url).to end_with root_path
+     end
+     it 'displays a helpful message' do
+       expect(page).to have_selector 'p.notice',
+         text: 'Stop -1 not found'
+     end
+   end
+   context 'without completing stop id' do
+     it 'autofills' do
```
jforgue: Is this functionality not complete yet?

spec/system/index_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+     end
+   end
+end
+
+describe 'searching for a bus stop by stop name' do
```
jforgue: Again this search doesn't seem super user friendly.

spec/system/index_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    it 'stays on the same page' do
+      expect(page.current_url).to end_with bus_stops_path
+    end
+    it 'displays a helpful message' do
+      expect(page).to have_selector 'p.notice',
+        text: 'Stop 1234 not found'
+    end
+  end
+  context 'without completing stop id' do
+    it 'autofills' do
```
jforgue: bump

spec/system/index_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+  end
+end
+
+describe 'searching for a bus stop by route' do
+  let(:user) { create :user }
+  let!(:route){ create :route }
+  let!(:bus_stop){ create :bus_stop, routes: [route] }
+  before :each do
```
jforgue: bump

spec/system/users/creating_users_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  context 'with all fields' do
+    before :each do
+      expect(page).to have_text 'New User'
+      within 'form#new_user.new_user' do
+        fill_in 'Name', with: 'Ben K'
+        fill_in 'Email', with: 'ben@example.com'
+        check 'Admin'
+        fill_in 'Password', with: 'password'
+        fill_in 'Password confirmation', with: 'password'
+
```
jforgue: Is this convention?

spec/system/users/creating_users_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+          fill_in 'Password confirmation', with: 'password7'
+
+          click_on 'Save user'
+        end
+     end
+     it 'sends a helpful error message' do
+       expect(page).to have_selector 'p.errors',
+         text: 'Password confirmation does not match password'
+     end
+     it 'redirects to edit user page' do
```
jforgue: It redirects to edit, does that mean it created the user even though the passwords didn't match?

config/initializers/paper_trail.rb
```diff
@@ -0,0 +1 @@
+PaperTrail.config.track_associations = false
```
jforgue: Just a question. Why don't we want to track associations?

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+describe 'managing stops as an admin' do
+  let(:admin) { create :user, :admin }
+  let!(:bus_stop) { create :bus_stop }
+  before :each do
+    when_current_user_is admin
+    visit manage_bus_stops_url
+  end
+  context 'delete button' do
+    it 'deletes the specific bus stop' do
+      expect(page).to have_selector 'table.manage tbody tr', count: 1
```
jforgue: What if there were more than one `bus_stop` on this page? Would it delete the right objet?

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+      page.driver.browser.switch_to.alert.accept
+      expect(page).to have_selector 'p.notice',
+        text: "#{bus_stop.name} has been deleted."
+      expect(page).not_to have_selector 'table.manage tbody tr', text: bus_stop.name
+    end
+  end
+  context 'edit button' do
+    it 'redirects to edit bus stop page' do
+      expect(page).to have_selector 'table.manage tbody tr', count: 1
```
jforgue: Same as previous comment.

spec/system/index_spec.rb
jforgue: Whops! Didn't see your comment on this.

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: As long as the _objet_ doesn't have the same name as another.

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: You're right, will fix.

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: @dfaulken mentioned something similar, but said that I didn't need to unless the line appeared multiple times. But if you think it's absolutely necessary I will make the appropriate changes.

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: You didn't stop commenting. Liar.

spec/system/users/creating_users_spec.rb
mboneil10: Is what convention?

mboneil10: Oh. Got it.

config/initializers/paper_trail.rb
mboneil10: Don't know

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'spec_helper'
+
+describe 'managing stops as an admin' do
+  let(:admin) { create :user, :admin }
+  let!(:bus_stop) { create :bus_stop }
+  before :each do
```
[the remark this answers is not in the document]
mboneil10: I disagree

spec/system/bus_stops/managing_bus_stops_spec.rb
dfaulken: (Bien que j'aimerais croire que @jforgue essayait de parler français, je crois qu'il a fait une erreur typographique.)

spec/system/bus_stops/managing_bus_stops_spec.rb
Anbranin: I disagree. There is no reason to not have both unless she were using a mix of `let` variables _and_ instance variables. The reason there isn't functional, it's decreasing confusion upon reading it. If she had a `@stop` and a `let :stop do create :stop` it would make any future developers pause and ask "wait, why? what's the difference?" And if there is none, that I'd say is a reason for refactor. But, she is not, it just _looks_ bad to you, but some people just would prefer to be consistent and use `let` across the board rather than futz with instance variables. I say "let" them. hahaha

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    expect(page).not_to have_selector 'table.manage tbody tr', text: format_time(present_stop.updated_at)
+  end
+  it 'allows editing of outdated stops' do
+    within 'tr', text: format_time(old_stop_1.updated_at) do
+      click_link 'Edit'
+    end
+    expect(page).to have_content "Editing #{old_stop_1.name}"
+  end
+  it 'outdated can be narrowed down with a different date' do
+    pending('review datepickers and capybara')
```
Anbranin: Ah, remember we had to click datepickers twice to get them to work?

spec/system/index_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+  context 'correct stop id' do
+    it 'redirects to the edit page' do
+      within 'form', text: 'Enter stop ID' do
+        fill_in 'Enter stop ID', with: bus_stop.hastus_id.to_s
+        click_button 'Search'
+      end
+      expect(page).to have_content "Editing #{bus_stop.name}"
+    end
+  end
+   context 'incorrect stop id' do
```
Anbranin: Is this indented one character over..?

spec/system/index_spec.rb
Anbranin: No it doesn't, but that's not what this PR is about

spec/system/index_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+        text: 'Stop 1234 not found'
+    end
+  end
+  context 'without completing stop name' do
+    let!(:bus_stop) { create :bus_stop, name: 'Mill Valley Apartments' }
+    it 'autofills' do
+      within 'form', text: 'Enter stop name' do
+        fill_in 'Enter stop name', with: 'Mill Valley'
+        choose_autocomplete_result 'Mill Valley Apartments', '#Enter stop name'
+        binding.pry
```
Anbranin: I'm assuming since isn't complete you don't want approval?

Anbranin: Also, are you having problems with this? Do you maybe need to `wait` a minute to let the autocomplete populate?

spec/system/index_spec.rb
Anbranin: no

spec/system/index_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+        click_button 'Search'
+      end
+    end
+  end
+end
+
+describe 'searching for a bus stop by route' do
+  let(:user) { create :user }
+  let!(:route){ create :route }
+  let!(:bus_stop){ create :bus_stop, routes: [route] }
```
Anbranin: Is this test working out for you OK?

spec/system/users/creating_users_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+   end
+   context "password and password_confirmation don't match" do
+     before :each do
+        within 'form#new_user.new_user' do
+          check 'Admin'
+          fill_in 'Name', with: 'Brody'
+          fill_in 'Email', with: 'brody@example.com'
+          fill_in 'Password', with: 'password'
+          fill_in 'Password confirmation', with: 'password7'
+
```
Anbranin: Why these spaces?

spec/system/users/editing_users_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      expect(page).to have_selector 'p.notice',
+        text: 'User was updated.'
+    end
+    it 'redirect to users table' do
+      expect(page.current_url).to end_with users_path
+    end
+  end
+  context 'with name, email, or admin changes' do
+    it 'updates the user and table accordingly' do
+      @previous_name = edit_user.name
```
Anbranin: These don't need to be instance variables do they?

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: Yeah but the calendar pops up and covers the field we'd like to `fill_in`

spec/system/index_spec.rb
mboneil10: Probably

spec/system/index_spec.rb
mboneil10: oh shit committed a binding. I'm officially a member of the club.

And yeah. Thanks @Anbranin

spec/system/index_spec.rb
mboneil10: ? Yes, why?

spec/system/users/creating_users_spec.rb
mboneil10: I was being inconsistent. 👍 will fix

spec/system/bus_stops/managing_bus_stops_spec.rb
dfaulken: But we don't want to fill it in. We want to use the datepicker.

spec/factories/bus_stops.rb
dfaulken: Okay, so sounds like we should allow for now.

spec/system/users/creating_users_spec.rb
mboneil10: Yes

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: @jforgue you spelled my name wrong. And I agree, I shouldn't rely on names for this test.

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: Ok, will do

spec/system/users/creating_users_spec.rb
jforgue: Is that the desired behavior?

jforgue: Nvm, I now understand that users can't access their account with no password.

spec/system/bus_stops/managing_bus_stops_spec.rb
dfaulken: Right, you can use `let` blocks to define objects because you prefer how they work to instance variables, and yet still have some actions that you need to perform before each test (logging in, visiting the page, etc.). Nothing wrong with that.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 21.72
+    "covered_percent": 83.44
```
mboneil10: I promise this is what it looked like on my local machine. Idk why it was getting pushed as 21.72.

PR #48, opened 2018-06-05
Completed only when Stop is Completed

### LINKED ISSUE #37: Only allow Completed Stop after All Fields are Filled

mboneil10:
Closes #37

### DIFF
--- a/app/models/bus_stop.rb
+++ b/app/models/bus_stop.rb
@@ -6,6 +6,7 @@ class BusStop < ApplicationRecord
   has_paper_trail
   validates :name, presence: true
   validates :hastus_id, presence: true, uniqueness: true
+  validates :check_completed, presence: { if: :completed? }
   has_and_belongs_to_many :routes
 
   before_save :assign_completion_timestamp, if: -> { completed_changed? }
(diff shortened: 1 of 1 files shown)


app/models/bus_stop.rb
```diff
@@ -6,6 +6,7 @@ class BusStop < ApplicationRecord
   has_paper_trail
   validates :name, presence: true
   validates :hastus_id, presence: true, uniqueness: true
+  validates :check_completed, presence: { if: :completed? }
```
dfaulken: To use a custom validation method, you want `validate`, not `validates`, correct?

I think it would be good to see a unit test in any case.

app/models/bus_stop.rb
```diff
@@ -83,4 +84,7 @@ def assign_completion_timestamp
     assign_attributes completed_at: (completed? ? DateTime.current : nil)
   end
 
+  def check_completed
+    errors.add :base, 'You must complete all fields before marking stop as complete.'
```
dfaulken: So this always adds an error for any completed stop?

I'm confused as to how there isn't any statement that checks for all of the completed fields in this PR.

mboneil10:
Closing; see #64

PR #56, opened 2018-06-06
Fix migrations for rails 5.1

jforgue:


db/migrate/20180319230634_add_sign_type_to_bus_stops.rb
```diff
@@ -1,4 +1,4 @@
-class AddSignTypeToBusStops < ActiveRecord::Migration[4.2]
```
Anbranin: The number here should be the rails version that this migration was written for.

Since there are no syntactical changes between migrations written for 4.2 and those written for 5.1 (that I know of.. at least not with these simple ones) changing this number becomes merely unnecessary and not harmful. I just wanted you to know.

PR #57, opened 2018-06-07
Order stop fields

### LINKED ISSUE #54: Order Stop Fields according to Field Guide
Requested during the May 31st PVTA meeting

jforgue:
Closes #54

### DIFF
--- a/app/models/bus_stop.rb
+++ b/app/models/bus_stop.rb
@@ -1,8 +1,9 @@
+# frozen_string_literal: true
+
 require 'csv'
 
-include DateAndTimeMethods
-
 class BusStop < ApplicationRecord
+  include DateAndTimeMethods
   has_paper_trail
   validates :name, presence: true
   validates :hastus_id, presence: true, uniqueness: true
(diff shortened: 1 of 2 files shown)


jforgue:
I didn't add any missing fields or remove unnecessary fields from the field guid here via migration. However, I did opt not to use the extra field in the view.

dfaulken:
May we see the final product in action?

app/models/bus_stop.rb
```diff
@@ -1,48 +1,91 @@
...
+    shelter_options: SHELTER_OPTIONS,
+    amenities: AMENITIES,
+    accessibility: ACCESSIBILITY,
+    security_and_safety: SECURITY_AND_SAFETY,
+    technology: TECHNOLOGY
+  }.freeze
+
+  SUPER_HASH.each do |hash|
+    hash.each do |name, options|
+      if options.kind_of?(Array)
```
dfaulken: Would not `is_a?` work here?

jforgue: It would, is that preferred?

jforgue:
screen shot 2018-06-07 at 10 57 14 am

dfaulken:
Cool. We can quibble about styling more when we're switching to the long term solution.

app/models/bus_stop.rb
dfaulken: I guess if it quacks like an array, we can use it.

werebus: Of course, if you really wanted to take duck-typing to its extreme, you would do:

```ruby
if options.respond_to? :include?
```

PR #58, opened 2018-06-07
Rewrite

mboneil10:
Halfway through the model and stop app rewrite


db/seeds.rb
```diff
@@ -12,6 +12,25 @@
              password_confirmation: 'password',
              admin: false
 
+Field.create! name: 'Curb cut',
+              category: 'Accessible',
+              field_type: :choice,
+              #choices: ["Within 20'", 'No curb cut', 'No curb'],
```
mboneil10: I didn't know how to include this in fields. :/

db/migrate/20180606143538_create_bus_stop_fields.rb
```diff
@@ -0,0 +1,10 @@
+class CreateBusStopFields < ActiveRecord::Migration[5.1]
+  def change
+    create_table :bus_stop_fields do |t|
+      t.string :value
+
+      t.timestamps
+    end
+    create_join_table :bus_stops, :fields
```
dfaulken: Does this not attempt to create the table twice? I would have expected just a singular `create_table`, since `create_join_table` doesn't allow specifying extra fields.

db/schema.rb
```diff
@@ -48,11 +54,26 @@
     t.string "shelter_type"
   end
 
+  create_table "bus_stops_field_names", id: false, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
dfaulken: This seems like an accidental inclusion?

mboneil10: tis

db/migrate/20180606143538_create_bus_stop_fields.rb
mboneil10: I found this online: 
[`create_join_table(table_1, table_2, options = {}) public`](https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/create_join_table)
That's why I put two fields down. Am I misinterpreting what the two tables are?

db/seeds.rb
dfaulken: Text, and then serialize to an Array.

lib/tasks/bus_stop_field.rb
```diff
@@ -0,0 +1,68 @@
...
+      ],
+      'Amenity' => [
+        { name: 'Bench', column: :bench, field_type: :choice, choices: ['PVTA bench', 'Other structure', 'Other bench', 'PVTA and other bench', 'None'], description: 'Determine if the bus stop has a bench or area to sit.' },
+        { name: 'Bike rack', column: :bike_rack, field_type: :choice, choices: ['PVTA bike rack', 'Other bike rack', 'PVTA and other bike rack', 'None'], description: 'Determine if there are bike racks.' },
+        { name: 'Schedule holder', column: :schedule_holder, field_type: :choice, choices: ['On pole', 'In shelter', 'None'], description: 'Determine if there is a bus schedule or not.' },
+        { name: 'System map exists', column: :system_map_exists, field_type: :choice, choices: ['New map', 'Old map', 'None'], description: 'Determine if there is a system map at this stop.' },
+        { name: 'Trash', column: :trash, field_type: :boolean, description: "Determine if there is a trash receptacle at the stop or within 10' of the stop." }
+      ],
+      'Accessible' => [
+        { name: 'Accessible', column: :accessible, field_type: :boolean, description: 'Can this stop be safely accessed?' },
```
dfaulken: Personally, I would align each field vertically, but the tradeoff is line length, so it's up to you.

lib/tasks/bus_stop_field.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    fields.each_pair do |category, category_fields|
+      category_fields.each.with_index do |category_field, rank|
+        field = Field.create!(name: category_field.fetch(:name),
+                              rank: rank,
+                              category: category,
+                              description: category_field.fetch(:description),
+                              field_type: category_field.fetch(:field_type))
+        BusStop.all.each do |stop|
+          BusStopField.create! stop: stop, field: field, value: stop.send(category_field.fetch(:column))
```
dfaulken: Are there any fields that don't currently have a column in the bus stop table?

If there are, what would happen here?

db/migrate/20180606143538_create_bus_stop_fields.rb
dfaulken: Do you want this migration to create two tables, though? My reading is that the first `create_table` makes a table called `bus_stop_fields`, and the `create_join_table` creates a table called `bus_stops_fields`, and with no `value` column.

Neither of those seems correct to me, but in either event I assume you don't want both.

db/seeds.rb
```diff
@@ -12,6 +12,25 @@
...
+              description: 'Determine the border between the road and the bus stop.',
+              rank: 2
```
dfaulken: I don't see a rank 1 'Accessible' field.

lib/tasks/bus_stop_field.rb
```diff
@@ -0,0 +1,68 @@
...
+        { name: 'Solar lighting', column: :solar_lighting, field_type: :boolean, description: 'Determine if this stop has solar lighting.' },
+        { name: 'Has power', column: :has_power, field_type: :choice, choices: ['Yes stub up', 'Yes outlet', 'No'], description: 'Determine if there is power or not.' },
+        { name: 'Real time information', column: :real_time_information, field_type: :choice, choices: ['Yes - solar', 'Yes - power', 'No'], description: 'Determine if this location has a real time information sign.' }
+      ]
+    }
+
+    fields.each_pair do |category, category_fields|
+      category_fields.each.with_index do |category_field, rank|
+        field = Field.create!(name: category_field.fetch(:name),
+                              rank: rank,
```
dfaulken: Have you run this task? Does it work?

Is rank `0` a problem with your design?

db/migrate/20180606143538_create_bus_stop_fields.rb
mboneil10: AH. I get what you mean. Thanks. I'll work on it tomorrow

db/seeds.rb
mboneil10: ? What do you mean.

lib/tasks/bus_stop_field.rb
mboneil10: I was considering it. Perhaps I will.

lib/tasks/bus_stop_field.rb
mboneil10: I haven't. And oh it certainly is a problem with my design.

dfaulken: Sorry, let me clarify the question. Would it cause a problem, given your design?

lib/tasks/bus_stop_field.rb
mboneil10: value would be nil...?

db/seeds.rb
dfaulken: This is a rank 2 field, but I didn't see a rank 1 field.

lib/tasks/bus_stop_field.rb
dfaulken: I suspect you'd be sending a method to which the model does not respond.

mboneil10: OK

db/migrate/20180606143538_create_bus_stop_fields.rb
mboneil10: Actually, I don't understand why the first table is wrong.

db/seeds.rb
mboneil10: OH WHOOPS my bad

db/migrate/20180606143538_create_bus_stop_fields.rb
```diff
@@ -0,0 +1,9 @@
+class CreateBusStopFields < ActiveRecord::Migration[5.1]
+  def change
+    create_table :bus_stop_fields do |t|
```
mboneil10: To continue our discussion @dfaulken, why is this wrong?

mboneil10: I'm not fighting for this being right, I just don't know why it's wrong.

db/migrate/20180606143538_create_bus_stop_fields.rb
dfaulken: Typically join tables would have the pluralized names of both tables. As you've written it, `bus_stop` is singular.

lib/tasks/bus_stop_field.rb
mboneil10: It works, but rank 0 doesn't make too much sense.

dfaulken: Keep in mind that array indices start at 0, is why I bring it up.

The value isn't constrained in the model, so arguably I could set negative rank as well.

db/migrate/20180606143538_create_bus_stop_fields.rb
dfaulken: It's not a big deal since you're using an Active record model for this anyway. Conventionally join tables will be the two table names, both pluralized, in alphabetical order.

But since the model is called BusStopField, arguably it's also convention for the table to be named bus stop fields.

So just removing the create join table statement is probably fine.

lib/tasks/bus_stop_field.rb
mboneil10: That's true.

lib/tasks/bus_stop_field.rb
mboneil10: No method error

mboneil10: But the value should be nil. SO I need to make some changes.

lib/tasks/bus_stop_field.rb
mboneil10: I was going to get to model constraints later, but just to be sure I should probably get them done now.

mboneil10:
Welp. I'm going to try running this now.

mboneil10:
It doesn't work.

lib/tasks/bus_stop_field.rb
```diff
@@ -0,0 +1,288 @@
...
+      category_fields.each.with_index do |category_field, rank|
+        rank += 1
+        field = Field.create! name: category_field.fetch(:name),
+                              rank: rank,
+                              category: category,
+                              description: category_field.fetch(:description),
+                              field_type: category_field.fetch(:field_type),
+                              choices: category_field.fetch(:choices)
+        BusStop.all.each do |stop|
+          BusStopField.create! bus_stop: stop,
```
mboneil10: `bus_stop` and `field` are missing in `BusStopField`.

mboneil10: oh right. Nested attributes

mboneil10: Damn ok I'll try some stuff

dfaulken: I don't understand what you're referring to here.

Is there a problem I can help diagnose?

mboneil10: Yes. I wrote the migration incorrectly. Now that I've written it correctly (a BusStopField has the attributes `bus_stop` and `field`), the join table is connected by the `bus_stop_id` and `field_id`. I thought that adding a foreign key of `field_name` would solve that problem, but it doesn't.

mboneil10:
@anbranin convinced me that we shouldn't be joining on `field_name` and `id`. 

The reason we wanted to join on `field_name` and `bus_stop_id` was for the scenario that a `field_name` was updated. Then, any of the data in HASTUS would still link to that `field`, even though it was completely different. HOWEVER:

When would a user ever want to change _just_ the name of the `field`and have it be a completely different question? Would they not need to also change the corresponding `description`? And, possibly `field_type`? At that point, the `field` shouldn't be updated and one could just as easily add a new `field`.

mboneil10:
Alright, I'll figure out what's wrong with the rake task on my own and run it. Thanks all :)

mboneil10:
LOL I had the rake file end in `rb`.

dfaulken:
No, I'm going to reclose. GH is showing things in the diff that are already live on master.

Issue #20, opened 2018-03-16
Date of stop survey


sherson:
We track when a stop was edited, but we should also probably track when it was actually physically checked/inventoried, and by whom. Related to #18.

dfaulken:
Should this be done automatically, or by manual user input (perhaps a separate model who tracks what stops were surveyed when, with a text field where users can indicate what person/group/entity/agency surveyed it)?

sherson:
All good questions. I'd like to talk with the other garages and PVTA first to get their thoughts.

mboneil10:
During the PVTA meeting on May 31st, it was requested that a user should be able to verify they physically checked/inventoried a stop. @sherson and I decided a button is best.

dfaulken:
A button that does what? Will there be a separate timestamp for it? Will we track who pressed it?

sherson:
>A button that does what? Will there be a separate timestamp for it? Will we track who pressed it?

It will track who pressed it and when.

mboneil10:
Do we think this should be on the edit page or on the manage stops page? Or both?

mboneil10:
After a meeting on June 5, we're going back to the idea of figuring out when the check box is checked. So, using paper trail if we can.

sherson:
I was thinking it should be on the edit page

sherson:
I was asked about this. To clarify, we should track both when it was last checked, and by whom (not be confused with when a stop was modified on a computer days / weeks / months later). That will come in handy for PVTA during the inventory process, and possibly years later.

PR #59, opened 2018-06-11
Add missing fields and remove unused field

jforgue:


jforgue:
screen shot 2018-06-11 at 3 41 25 pm

Looks more or less the same ass before.

app/views/bus_stops/_form.haml
```diff
@@ -21,6 +18,18 @@
...
         %td= f.check_box :completed
+      %tr
+        %td= f.label :garage_responsible
+        %td= f.select :garage_responsible, %w[SATCo UMTS VATCo], include_blank: true
+      %tr
+        %td= f.label :state_road
+        %td= f.check_box :state_road
+      %tr
+        %td= f.label :date_stop_checked
+        %td= f.date_select :date_stop_checked, value: Date.today, disabled: true
```
dfaulken: @mboneil10 my interpretation is that these ought to be autofilled by the controller when (and only when) someone is marking the stop as completed.

Is that your understanding?

mboneil10: Yes, my fault I forgot to tell him to look at that issue.

mboneil10: @jforgue, in #20 we concluded to do what @dfaulken said above. You can remove `:date_stop_checked` and `:stop_checked_by`. Sorry I didn't realize the confusion earlier.

jforgue: Will do, no worries.

jforgue: Do we just want the fields removed from the form or the attributes removed from the model? I assume the later as we mention `paper_trail` in the issue.

Anbranin: As we talked about, both of these attributes can be covered with PaperTrail `versions` and `whodunnit`, so I'd remove the columns entirely at least for now.

dfaulken: We'd discussed that that could have performance implications, so I agree to use PaperTrail for now, but we should keep the current approach in our back pocket.

PR #64, opened 2018-06-14
Add completed by

### LINKED ISSUE #20: Date of stop survey

Anbranin:
~I didn't use an ActiveRecord relation because I thought a bus stop belonging to a user was too weird.. it doesn't seem like the right kind of relationship. Then again, perhaps I'm thinking about OOP in the wrong way. Let me know, it's easy to add that relationship.~ I did end up using an ActiveRecord Relation.

closes #20


app/models/bus_stop.rb
```diff
@@ -182,9 +182,17 @@ def self.to_csv(limited_attributes: false)
...
+    User.find(completed_by).try :name
+  end
+
   private
 
-  def assign_completion_timestamp
-    assign_attributes completed_at: (completed? ? DateTime.current : nil)
+  def assign_completion_attributes
+    whodunnit = versions.last.whodunnit
+    assign_attributes(
```
dfaulken: I think it reads more easily with the conditional a level up, and using setters rather than `assign_attributes`:

```ruby
if completed?
  completed_at = DateTime.current
  completed_by = versions.last.whodunnit
else
  completed_at = nil
  completed_by = nil
end
```

Anbranin: It seemed cleaner the other way but I've always enjoyed putting ternary operators everywhere I could. I toyed with this idea though and feel neutrally. Will change.

I'll see about adding the relationship.

Anbranin: Ooh this is a gotcha:
`completed_at = DateTime.current`
- Assigns a local variable with the value DateTime.current.
`self.completed_at = DateTime.current`
- Does the thing you want

dfaulken: If so, then that's a serious bug in Rails. Setter methods are a 100% valid way of setting attributes in Rails.

jforgue: Personally I like the ternaries, I think it looks cleaner. I know @mboneil10 would disagree though!

app/controllers/bus_stops_controller.rb
```diff
@@ -64,6 +64,7 @@ def outdated
 
   def update
     @stop.assign_attributes stop_params
+    @stop.completed? ? @stop.completed_by = current_user : nil
```
dfaulken: Two issues with this:

1. This is logically equivalent to `@stop_completed.by = current_user if @stop.completed?`, because your else case does nothing.
2. It doesn't handle if the stop was already completed.

app/controllers/bus_stops_controller.rb
```diff
@@ -64,6 +64,7 @@ def outdated
 
   def update
     @stop.assign_attributes stop_params
+    @stop.assign_completed_by current_user
```
dfaulken: Style comment, slightly misleading in my opinion  to say "assign completed by current user" when the majority of the time, we don't actually do what those words suggest.

app/models/bus_stop.rb
```diff
@@ -7,6 +7,7 @@ class BusStop < ApplicationRecord
   has_paper_trail
   validates :name, presence: true
   validates :hastus_id, presence: true, uniqueness: true
+  belongs_to :completed_by, class_name: 'User', foreign_key: 'completed_by'
```
dfaulken: I believe the foreign key can (and arguably should) be a symbol.

mboneil10:
Closes #37

Anbranin:
For posterity: jforgue deleted a comment from @mboneil10 that was mistakenly saying that this closes issue thirty-seven. It does not.

PR #66, opened 2018-06-14
Rewrite

dfaulken:
Much of this content comes from #58.

My addition is redefining the association between fields and bus stop fields, as well as fixing and cleaning up the rake task. See discussion on Slack for why I maintain that this is the preferable way to define the association.

Confirmed in a Rails console that all associations are functional and correct, and that the rake task runs properly.

Once this is reviewed, we should close *without merging*, and then go to work on the controllers and views.

### DIFF
--- a/app/models/bus_stop_field.rb
+++ b/app/models/bus_stop_field.rb
@@ -1,4 +1,4 @@
 class BusStopField < ApplicationRecord
   belongs_to :bus_stop
-  belongs_to :field, primary_key: :name, foreign_key: :field_name
+  belongs_to :field, foreign_key: :field_name
 end
--- a/app/models/bus_stop_field.rb
+++ b/app/models/bus_stop_field.rb
@@ -1,0 +1,4 @@
+class BusStopField < ApplicationRecord
+  belongs_to :bus_stop
+  belongs_to :field
+end
--- a/app/models/bus_stop_field.rb
+++ b/app/models/bus_stop_field.rb
@@ -1,4 +1,4 @@
 class BusStopField < ApplicationRecord
   belongs_to :bus_stop
-  belongs_to :field
+  belongs_to :field, primary_key: :name, foreign_key: :field_name
 end
--- a/app/models/bus_stop_field.rb
+++ b/app/models/bus_stop_field.rb
@@ -1,4 +1,13 @@
 class BusStopField < ApplicationRecord
   belongs_to :bus_stop
   belongs_to :field, foreign_key: :field_name
+  
+  validate :check_choice_included
+  
+  private
+    def check_choice_included
+      if field.multiple_choice? && !field.choices.include?(value)
+        errors.add :value, 'is not a valid choice for this field'
+      end
+    end
 end
--- a/db/migrate/20160708192840_add_description_to_routes.rb
+++ b/db/migrate/20160708192840_add_description_to_routes.rb
@@ -1,8 +1,4 @@
-<<<<<<< HEAD
-class AddDescriptionToRoutes < ActiveRecord::Migration[5.1]
-=======
 class AddDescriptionToRoutes < ActiveRecord::Migration[4.2]
->>>>>>> c9726e8808c30c9a3f35e81b393f06d088a33726
   def change
     add_column :routes, :description, :string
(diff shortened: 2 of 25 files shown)


db/seeds.rb
```diff
@@ -12,6 +12,25 @@
...
+              field_type: :choice,
+              choices: ["Within 20'", 'No curb cut', 'No curb'],
+              description: 'Determine the border between the road and the bus stop.',
+              rank: 1
+
+Field.create! name: 'Solar lighting',
+              category: 'Technology',
+              field_type: :boolean,
+              description: 'Determine if this stop has solar lighting.',
+              rank: 1
```
Anbranin: What's a rank?

lib/tasks/create_bus_stop_fields.rake
```diff
@@ -0,0 +1,286 @@
+# frozen_string_literal: true
+
+namespace :bus_stop_fields do
+  task create: :environment do
+    fields = {
+      'Info' => [
+        {
+          name: 'Completed',
+          column: :completed,
```
Anbranin: What is "column" in this context?

db/seeds.rb
dfaulken: At what position within a particular category a field should fall.

lib/tasks/create_bus_stop_fields.rake
dfaulken: The bus stops table column from which the value should be pulled in order to populate the bus stop field.

db/migrate/20160708192840_add_description_to_routes.rb
```diff
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
 class AddDescriptionToRoutes < ActiveRecord::Migration[5.1]
+=======
+class AddDescriptionToRoutes < ActiveRecord::Migration[4.2]
+>>>>>>> c9726e8808c30c9a3f35e81b393f06d088a33726
```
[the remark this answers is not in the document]
dfaulken: Indeed. Thank you.

app/models/bus_stop_field.rb
```diff
@@ -0,0 +1,13 @@
+class BusStopField < ApplicationRecord
+  belongs_to :bus_stop
+  belongs_to :field, foreign_key: :field_name
+  
+  validate :check_choice_included
+  
+  private
+    def check_choice_included
+      if field.multiple_choice? && !field.choices.include?(value)
+        errors.add :value, 'is not a valid choice for this field'
```
jforgue: Is the purpose of this to ensure the response matches with Hastus?

dfaulken: Making sure that the value is one of the possible choices.

jforgue: Right. I guess this is kind of a moot point since I don't disagree with this validation. Does Hastus really care what string we send it? That's my real question. You would probably have to be selecting a different answer intentionally to get something outside of the options.

dfaulken: We are considering how to implement a Hastus-side validation that a choice is in a particular list.

db/seeds.rb
Anbranin: oh! cool

lib/tasks/create_bus_stop_fields.rake
Anbranin: Ohhh

db/migrate/20180614232518_drop_fields.rb
```diff
@@ -0,0 +1,5 @@
+class DropFields < ActiveRecord::Migration[5.1]
+  def change
+    drop_table :fields
```
jforgue: Does this have to be dropped in order to make the joins table?

dfaulken: See the following migration. I re-added the fields table, but without ID.

jforgue: Nvm I see you drop it to recreate with `id: false`.

dfaulken:
Thank you both. I will leave open for now. We should close before working on the views and controllers.

PR #69, opened 2018-06-16
Rewrite edit view

dfaulken:
How does this seem as a way of setting up the edit page?

*Should merge* once approved, as the base is `rewrite`. Then the final step is the controller.

### DIFF
--- a/app/controllers/bus_stops_controller.rb
+++ b/app/controllers/bus_stops_controller.rb
@@ -77,6 +77,7 @@ class BusStopsController < ApplicationController
 
   def find_stop
     @stop = BusStop.find_by hastus_id: params.require(:id)
+    @data_fields = @stop.data_fields
     unless @stop.present?
       redirect_back(fallback_location: root_path,
         notice: "Stop #{params[:id]} not found") and return
--- a/app/models/bus_stop.rb
+++ b/app/models/bus_stop.rb
@@ -1,155 +1,32 @@
 # frozen_string_literal: true
 
 require 'csv'
+include DateAndTimeMethods
 
 class BusStop < ApplicationRecord
-  include DateAndTimeMethods
   has_paper_trail
-  validates :name, presence: true
-  validates :hastus_id, presence: true, uniqueness: true
+
   has_and_belongs_to_many :routes
   has_many :bus_stop_fields
   has_many :fields, through: :bus_stop_fields
+
+  accepts_nested_attributes_for :bus_stop_fields
+
+  validates :name, presence: true
+  validates :hastus_id, presence: true, uniqueness: true
 
   before_save :assign_completion_timestamp, if: -> { completed_changed? }
 
   scope :not_updated_since,
         ->(date) { where 'updated_at < ?', date.to_datetime }
 
-  SIGN_OPTIONS = {
-    sign_type: ['Axehead (2014+)',
-                'Rectangle (<2014)',
-                'MGM + Axhead (2018+)',
-                'Other',
-                'No sign'],
-    mounting: ['PVTA pole',
-               'Other pole',
-               'City pole',
-               'Utility pole',
-               'Structure',
-               'No sign'],
-    shared_sign_post: ['No',
-                       'Yes - Traffic sign',
-                       'Yes - FRTA',
-                       'Yes - Other',
-                       'Sign not on pole'],
-    mounting_direction: ['Towards street',
-                         'Away from street',
-                         'Center',
-                         'No sign'],
-    mounting_clearance: ['Less than 60 inches',
-                         '60-83 inches',
-                         'Greater than 84 inches',
-                         'No sign'],
-    bolt_on_base: :boolean,
-    stop_sticker: ['No sticker',
-                   'Sticker incorrect',
-                   'Sticker correct'],
-    route_stickers: ['No stickers',
-                     'Stickers incorrect',
-                     'Stickers correct']
-  }.freeze
-
-  SHELTER_OPTIONS = {
-    shelter: ['No shelter',
-              'PVTA shelter',
-              'Other',
-              'Building'],
-    shelter_type: ['Modern',
-                   'Modern half',
-                   'Victorian',
-                   'Dome',
-                   'Wooden',
-                   'Extra large',
-                   'Other',
-                   'No shelter'],
-    shelter_ada_compliance: :boolean,
-    shelter_condition: ['Great',
-                        'Good',
-                        'Fair',
-                        'Poor',
-                        'No shelter'],
-    shelter_pad_condition: ['Great',
-                            'Good',
-                            'Fair',
-                            'Poor',
-                            'No shelter'],
-    shelter_pad_material: ['Asphalt',
-                           'Concrete',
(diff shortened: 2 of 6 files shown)


dfaulken:
screen shot 2018-06-16 at 12 02 07 pm

app/models/bus_stop.rb
```diff
@@ -1,155 +1,32 @@
...
-  include DateAndTimeMethods
   has_paper_trail
-  validates :name, presence: true
-  validates :hastus_id, presence: true, uniqueness: true
+
   has_and_belongs_to_many :routes
   has_many :bus_stop_fields
   has_many :fields, through: :bus_stop_fields
 
+  accepts_nested_attributes_for :bus_stop_fields
```
dfaulken: This is the only real change in this section. Everything else I just moved around for clarity.

app/models/field.rb
```diff
@@ -6,12 +6,18 @@ class Field < ApplicationRecord
...
+
+  scope :in_category, ->(category) { where category: category }
+
   def multiple_choice?
     field_type == :choice
   end
+
+  def self.categories
+    order(:category).pluck(:category).uniq
+  end
```
dfaulken: These methods are only used from the bus stop model, but I still prefer keeping some of the field-related functionality in its own model file.

app/views/bus_stops/_form.haml
```diff
@@ -6,37 +6,34 @@
...
+        %th
+          Stop Info
+    %tbody
       %tr
         %td= f.label :name
         %td= stop.name
       %tr
         %td= f.label :hastus_id, 'Hastus ID'
         %td= stop.hastus_id
-      - if current_user.admin?
```
dfaulken: I do not know why we would wish to only show the route list to admins. Potentially a holdover from when it was editable?

db/seeds.rb
```diff
@@ -63,7 +46,7 @@
 stops.each do |route_number, stop_names|
   stop_names.each do |stop_name|
     # Anytime in the last two months
-    Timecop.freeze rand(86_400).minutes.ago do
+    Timecop.freeze Time.now - rand(2.months) do
```
dfaulken: I saw this and had to wonder why I didn't just do it originally.

lib/tasks/create_bus_stop_fields.rake
```diff
@@ -268,6 +268,7 @@ namespace :bus_stop_fields do
       ]
     }
 
+    Field.delete_all
```
dfaulken: Just in case

mboneil10: Good call

app/models/bus_stop.rb
```diff
@@ -1,155 +1,32 @@
...
-    security_and_safety: SECURITY_AND_SAFETY,
-    technology: TECHNOLOGY
-  }.freeze
-
-  SUPER_HASH.each do |hash|
-    hash.each do |name, options|
-      if options.kind_of?(Array)
-        validates name, inclusion: { in: options }, allow_blank: true
+  def data_fields
+    Field.categories.map do |category|
```
mboneil10: Cool, this whole method is good.

app/models/field.rb
```diff
@@ -6,12 +6,18 @@ class Field < ApplicationRecord
...
   validates :category, presence: true
   validates :rank, presence: true, uniqueness: { scope: :category }
-  validates :field_type, inclusion: { in: %i[boolean choice] }
+  validates :field_type, inclusion: { in: %w[boolean choice] }
   
   serialize :choices, Array
   validates :choices, presence: true, if: :multiple_choice?
-  
+
+  scope :in_category, ->(category) { where category: category }
```
mboneil10: 👍

app/models/field.rb
mboneil10: I agree with that preference.

app/views/bus_stops/_form.haml
```diff
@@ -6,37 +6,34 @@
...
-              - else
-                = f.check_box name
+      - @data_fields.each_pair do |category, category_fields|
+        %tr
+          %th{ colspan: 2 }= category
+        - category_fields.each do |data_field|
+          = fields_for data_field do |data_field_form|
+            %tr
+              %td= data_field_form.label :value, data_field.field_name
+              - case data_field.field.field_type
```
mboneil10: Hooray!

db/seeds.rb
```diff
@@ -12,24 +12,7 @@
...
-              field_type: :boolean,
-              description: 'Determine if this stop has solar lighting.',
-              rank: 1
-
-Field.create! name: 'Completed',
-              category: 'Info',
-              field_type: :boolean,
-              description: 'Have all attributes been checked?',
-              rank: 1
+# Recommended to create fields with `rails bus_stop_fields:create`.
```
mboneil10: Ok

app/views/bus_stops/_form.haml
dfaulken: Definitely very open to suggestions for better local variable names.

db/seeds.rb
dfaulken: Seems like we wouldn't want a different set of fields in dev vs. prod, similar to how right now we don't have a different set of columns on the bus stop table.

mboneil10: Got i

app/views/bus_stops/_form.haml
mboneil10: Too many `field`s for you?

app/controllers/bus_stops_controller.rb
```diff
@@ -77,6 +77,7 @@ def update
 
   def find_stop
     @stop = BusStop.find_by hastus_id: params.require(:id)
+    @data_fields = @stop.data_fields
```
Anbranin: I wouldn't have bothered assigning this to its own instance variable personally.

Anbranin: Actually reading further I see why you did it.

app/models/bus_stop.rb
```diff
@@ -172,16 +49,7 @@ def route_list
...
-              LIMITED_ATTRS
-            else Hash[columns.map { |c| [c.name, c.name.humanize] }]
-            end
-    CSV.generate headers: true do |csv|
-      csv << attrs.values
-      all.each do |stop|
-        csv << attrs.keys.map { |attr| stop.send attr }
-      end
-    end
+    # TODO
```
Anbranin: This is OK?

app/controllers/bus_stops_controller.rb
dfaulken: Collect your data in the controller, was my thinking.

app/models/bus_stop.rb
dfaulken: Out of scope for this PR, I think, but should definitely be done before we merge to master, agreed.

app/views/bus_stops/_form.haml
Anbranin: One app where we have to make a form for a model called "Field" is too many

Issue #68, opened 2018-06-15
Replace check boxes with select fields/radio buttons


jforgue:
It's been requested that these yes/no questions have a blank default so we can verify that the user has checked the field before completing the form.

sherson:
Right, radio buttons have a slight UI advantage (one tap instead of two), but either approach is fine.

jforgue:
@dfaulken My thoughts exactly. My understanding is that we want this issue solved for Wednesday. Considering that, if I were working on this issue I would go for select fields and move on to something more pressing.

sherson:
UMTS Service confirmed that the checkboxes for any fields for any stops that they marked as complete were updated. That is to say, for stops that UMTS marked as complete since 6/11/2018, anything that is unchecked is assumed to be `No` (while for stops that UMTS did not mark as complete since 6/11, unchecked is assumed to be blank).

sherson:
UMTS Service is not using the app again until Wednesday, so these changes can be made anytime today or tomorrow.

mboneil10:
I'm impartial to this. Radio buttons seem like the way to do this because then we still see that they are bools. I guess a user wouldn't care about that, though.

mboneil10:
And I think the UI advantage is enough for me to make them radio buttons.

dfaulken:
So how about the converse? Should all select fields be displayed as radio buttons? That way all options are readily available in front of you?

mboneil10:
Hm.

mboneil10:
@dfaulken are you playing devil's advocate? Dammit. I don't know what to do.

mboneil10:
Okay, idea is to try out both views: 1) all as select boxes, 2) some as select boxes, some as radio buttons. Then go from there.

jforgue:
Closed with #72.

PR #72, opened 2018-06-18
Check box

mboneil10:
First Way:
screen shot 2018-06-18 at 4 35 25 pm

### DIFF
--- a/app/views/bus_stops/_form.haml
+++ b/app/views/bus_stops/_form.haml
@@ -19,6 +19,10 @@
           %td= stop.route_list
       %tr
         %td= f.label :completed, 'Completed?'
+        %td= 'Yes'
+        %td= 'No'
+      %tr
+        %td
         %td= f.radio_button 'completed', 'yes'
         %td= f.radio_button 'completed', 'no'
       %tr
@@ -26,6 +30,10 @@
         %td= f.select :garage_responsible, %w[SATCo UMTS VATCo], include_blank: true
       %tr
         %td= f.label :state_road
+        %td= 'Yes'
+        %td= 'No'
+      %tr
+        %td
         %td= f.radio_button 'state_road', 'yes'
         %td= f.radio_button 'state_road', 'no'
     - BusStop::SUPER_HASH.each do |category, hash|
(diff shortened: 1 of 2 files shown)


mboneil10:
Just select boxes
screen shot 2018-06-18 at 4 41 44 pm

mboneil10:
Closes #68

mboneil10:
@anbranin suggested that I show the difference to a user (someone who works in service), since they might have an opinion on this. Perhaps a bit of bikeshedding on my part.

dfaulken:
Given that this will change how every field looks in the site's most important functionality, I would argue to run this by the group if feasible.

mboneil10:
@dfaulken what group?

dfaulken:
The project stakeholders. UMTS Service, PVTA, PVPC.

sherson:
> The project stakeholders. UMTS Service, PVTA, PVPC.

I already asked the group. Either is fine, so it's up to us. @mboneil10 said that it'd be slightly easier to go with select fields, so let's do that.

dfaulken:
Easier how? We noted that radio buttons require way fewer taps to enter a particular stop. What problems do we foresee with implementing radio buttons @mboneil?

jforgue:
Added a little styling in order to separate to complete check box moved to the bottom. Let me know what you think.

screen shot 2018-06-19 at 12 46 27 pm

Anbranin:
I don't like it, it's in a heading box but not a heading. Why not call it "Completed" even though that may look a little silly? There is also nothing wrong with including a _check when all fields are filled_ guy underneath it.

jforgue:
I'll fix the heading. What about the styling though? is this better or worse than what we had for organization?

Anbranin:
The question really is how does it look on a shrunken screen

jforgue:
Here is how the app looks on my phone: 

screenshot_2018-06-20-12-58-17
screenshot_2018-06-20-12-58-22

app/views/bus_stops/_form.haml
```diff
@@ -14,30 +14,30 @@
...
             %td= f.label name
-            %td
-              - if options.kind_of?(Array)
-                = f.select name, options, include_blank: true
-              - else
-                = f.check_box name
+            - if options.kind_of?(Array)
+              %td= f.select name, options, include_blank: true
+            - else
+              %td= f.select name, %w[Yes No], include_blank: true
```
dfaulken: Have we confirmed that this saves correctly? Does `"No"` coming over the wire translate to a false boolean value being saved?

jforgue: Good call, we didn't even change the field types!

jforgue: However, I have confirmed that `%td= f.select name, %w[Yes No], include_blank: true` saves correctly as a string for both yes and no.

jforgue: Or were we hoping the inverse was true? As in we wanted `"No"` to translate to `false`, because that did not work.

PR #73, opened 2018-06-19
Pending status

### LINKED ISSUE #29: Status of Pending for Stops
Requested during PVTA meeting on April 30th, 2018:

Show that an incomplete stop is pending after someone started servicing it.

jforgue:
Closes #29

### DIFF
--- a/app/views/bus_stops/by_route.haml
+++ b/app/views/bus_stops/by_route.haml
@@ -3,9 +3,14 @@
 - if @stops.empty?
   No stops on this route
 - else
-  Completed
-  - stop_list @stops.completed
-  Not Completed
-  - stop_list @stops.not_completed
-  Pending
-  - stop_list @stops.pending
+  - @stops_arr.each_pair do |status, stops|
+    - if status
+      Not Completed
+    -else
+      Completed
+    Stops
+    %ul
+    - stops.each do |stop|
+      %li
+        = link_to edit_bus_stop_path(stop.hastus_id) do
+          - stop.name
--- a/app/views/bus_stops/by_route.haml
+++ b/app/views/bus_stops/by_route.haml
@@ -4,12 +4,8 @@
   No stops on this route
 - else
   Completed
-  = @stops.completed
+  - stop_list @stops.completed
   Not Completed
-  = @stops.not_completed
-  Stops
-  %ul
-  - @stops.each do |stop|
-    %li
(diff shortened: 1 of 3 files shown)


dfaulken:
A few questions:

1. Why is it called arr if it's a hash?
2. Why are the keys symbols, if we're just going to have to translate them into strings in the view?

Good call on separating pending, completed, and not started. I think pending and not completed are the same thing conceptually, so we should be careful about distinguishing not started, started but not completed i.e. pending, and completed.

dfaulken:
Also, what order do we think we should be putting these three groups in? Should it be according to priority, that is, pending, not started, then completed? Or should it feel more chronological, that is, not started, pending, then completed?

jforgue:
I think ordering by priority would be more convenient for the user.

jforgue:
Here's a screenshot
screen shot 2018-06-19 at 4 42 49 pm

app/views/bus_stops/by_route.haml
```diff
@@ -3,14 +3,10 @@
...
   No stops on this route
 - else
-  - @stops.each_pair do |complete, stops|
-    - if complete
-      Completed
-    - else
-      Not Completed
-    Stops
+  - @stops_hash.each_pair do |status, stops|
+    = status.humanize + ' Stops'
```
dfaulken: But again, why do this translation? Why not just put the string that we want into the view as the keys in the hash?

app/views/bus_stops/by_route.haml
```diff
@@ -3,14 +3,10 @@
...
-      Not Completed
-    Stops
+  - @stops_hash.each_pair do |status, stops|
+    = status + ' Stops'
     %ul
     - stops.each do |stop|
       %li
         = link_to edit_bus_stop_path(stop.hastus_id) do
-          - stop.name
+          - stop.name
```
Anbranin: Could you fix your text editor so it doesn't remove newlines from the end of files? I'm so sick of seeing this everywhere.

dfaulken: I am a fan of reviewing with `?w=1` as the querystring.

werebus: :+1: Or, there's actually an option in the "Diff settings" menu now.

<sub>Or, you know http://www.dealwithit.org/</sub>

PR #74, opened 2018-06-19
Validate completion

### LINKED ISSUE #37: Only allow Completed Stop after All Fields are Filled

jforgue:
Closes #37. I'm not sure this is the best way to do this. Please leave suggestions!

### DIFF
--- a/app/models/bus_stop.rb
+++ b/app/models/bus_stop.rb
@@ -13,16 +13,20 @@ class BusStop < ApplicationRecord
 
   scope :not_updated_since,
         ->(date) { where 'updated_at < ?', date.to_datetime }
-  validates :name, :hastus_id, :bench, :curb_cut, :lighting, :mounting,
-            :mounting_direction, :schedule_holder, :shelter, :sidewalk_width,
-            :trash, :bolt_on_base, :bus_pull_out_exists, :has_power,
-            :solar_lighting, :system_map_exists, :mounting_clearance,
-            :created_at, :updated_at, :sign_type, :shelter_condition,
-            :shelter_pad_condition, :shelter_pad_material, :shelter_type,
-            :shared_sign_post, :shelter_ada_compliance, :garage_responsible,
-            :bike_rack, :ada_landing_pad, :real_time_information, :state_road,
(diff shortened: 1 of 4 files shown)


app/models/bus_stop.rb
```diff
@@ -13,6 +13,16 @@ class BusStop < ApplicationRecord
...
+  validates :name, :hastus_id, :bench, :curb_cut, :lighting, :mounting,
+            :mounting_direction, :schedule_holder, :shelter, :sidewalk_width,
+            :trash, :bolt_on_base, :bus_pull_out_exists, :has_power,
+            :solar_lighting, :system_map_exists, :mounting_clearance,
+            :created_at, :updated_at, :sign_type, :shelter_condition,
+            :shelter_pad_condition, :shelter_pad_material, :shelter_type,
+            :shared_sign_post, :shelter_ada_compliance, :garage_responsible,
+            :bike_rack, :ada_landing_pad, :real_time_information, :state_road,
+            :need_work, :obstructions, :accessible, :stop_sticker,
+            :route_stickers, presence: true, if: :completed?
```
dfaulken: I would put the `presence`, etc.. on a new line so that it stands out from the giant list.

Not a blocking change. This is temporary.

jforgue:
I will wait until #72 is merged to prevent issues with the checkboxes.

app/models/bus_stop.rb
werebus:
```ruby
required_for_completion = %i[name hastus_id bench etc]

validates *required_for_completion, presence: true, if: :completed?
```

dfaulken: Sure, could do. Saves some space, I suppose.

werebus: Yeah, I mean, it's all code golf, really. But what I like is that it puts all of the stuff that `validates` is doing into one line.  It "`validates` the `presence` of all the fields which are `required_for_completion` `if` the stop is `completed`".

dfaulken: Agreed. Separates the intent from the content.

PR #75, opened 2018-06-20
Route variants

### LINKED ISSUE #71: Modify rake tasks for stops import from Hastus
https://github.com/umts/stop-project/tree/master/lib/tasks 

Newest version of stops with routes:
https://umass.box.com/s/q8xfvpusdaoei3hpsp60xtjg6reo9meo 

Newest version of stops for stop project:
https://umass.box.com/s/midj9ghefi2yd64m7un7qgyl448nh3vh

mboneil10:
Closes #71


mboneil10:
A problem I've found is that this creates `bus_stops_routes` with `nil` sequences...

mboneil10:
Not all of them are `nil` though. Looking into it. Might be a problem with my logic

mboneil10:
Sorry I committed my example file. I'll make sure to remove it.

mboneil10:
Yes, I know I still need to:
- include direction in `BusStopsRoute`
- validate `sequence` is positive

working on that.

lib/tasks/route_import.rake
```diff
@@ -5,12 +5,72 @@ require 'csv'
...
         stop.routes << route
+
+        direction = row['direction']
+        variant = row['variant']
+        stop_id = row['stp_identifier']
+        rank = row['stop_variant_rank']
+
+        @route_hash[route] ||= {}
+        @route_hash[route][direction] ||= {}
+        @route_hash[route][direction][variant] ||= []
```
jforgue: Could you combine these?

mboneil10: Into what

mboneil10: Do you understand what's happening on each line?

mboneil10:
`ActiveRecord::RecordInvalid: Validation failed: Sequence can't be blank, Sequence is not a number, Direction can't be blank` when I run the rake task. This is great.

mboneil10:
I think I figured it out?????

mboneil10:
The last `bus_stop_route` in the list isn't getting created.

app/models/bus_stops_route.rb
```diff
@@ -0,0 +1,7 @@
+class BusStopsRoute < ApplicationRecord
+  belongs_to :bus_stop
+  belongs_to :route
+
+  validates :sequence, presence: true, numericality: { greater_than_or_equal_to: 1 }
```
dfaulken: Sequence should also be unique relative to route and direction, correct?

Also bus stop ID, should be unique relative to route and direction.

db/seeds.rb
```diff
@@ -41,14 +41,20 @@
...
     # Anytime in the last two months
     Timecop.freeze rand(86_400).minutes.ago do
+      # don't want the sequences to start with 0.
+      sequence = sequence + 1
       stop = BusStop.find_or_initialize_by name: stop_name
       stop.hastus_id = hastus_ids.fetch(stop_name)
-      stop.routes << routes.fetch(route_number)
       stop.save!
+      route = routes.fetch(route_number)
+      direction = directions.sample
```
dfaulken: That doesn't sound right. This would be a random smattering of integers between north and south.

We want north: 1,2,3,4, etc. and south: 1,2,3,4, etc.

lib/tasks/route_import.rake
```diff
@@ -5,12 +5,74 @@ require 'csv'
 namespace :routes do
   task :import, [:csv_file] => :environment do |_, args|
     Route.delete_all
+    BusStopsRoute.delete_all
+    @route_hash = {}
```
dfaulken: In general I'm curious why you used instance variables throughout.

lib/tasks/route_import.rake
```diff
@@ -5,12 +5,73 @@ require 'csv'
...
+        @route_hash[route][direction] ||= {}
+        @route_hash[route][direction][variant] ||= []
+        @route_hash[route][direction][variant] << { stop_id => rank }
+      end
+    end
+
+    @stop_list = []
+    @other_variants = []
+
+    @route_hash.each do |route, directions|
```
dfaulken: Procedurally I would separate "deciding what the eventual sequence should be based on various variants" and "creating the bus stop fields with that sequence" into two separate loops, probably with an intermediate data structure.

As it stands this isn't the easiest to follow.

lib/tasks/route_import.rake
```diff
@@ -5,12 +5,73 @@ require 'csv'
...
+
+        direction = row['direction']
+        variant = row['variant']
+        stop_id = row['stp_identifier']
+        rank = row['stop_variant_rank']
+
+        @route_hash[route] ||= {}
+        @route_hash[route][direction] ||= {}
+        @route_hash[route][direction][variant] ||= []
+        @route_hash[route][direction][variant] << { stop_id => rank }
```
dfaulken: What about if it were an array into which you added stop_id according to the array index of `rank`?

That is `some_array[rank.to_i] = stop_id`?

db/seeds.rb
mboneil10: ?????

lib/tasks/route_import.rake
mboneil10: Idk

db/seeds.rb
dfaulken: There should be a contiguous set of sequence numbers per direction, starting at 1. My reading is that this will create for example

Sequence | Direction
:--- | ---
1 | North
2 | South
3 | South
4 | North
5 | North
6 | South
7 | North

etc.

lib/tasks/route_import.rake
mboneil10: Ah

dfaulken:
I would have a loop where I look through the data to decide what the
sequence of stops should be in each direction.

Then I would have another loop to create them according to each sequence.

dfaulken:
There's just local scope here, right? Just the one file, no classes or
modules at hand?

Local scope would be fine if so.

dfaulken:
Sorry, thoughts those comments would apply to the comment thread I replied to. On train.

lib/tasks/route_import.rake
mboneil10: OHH

mboneil10: So you're saying I should organize the stops per direction. And then create them.

mboneil10: Did it

app/models/bus_stops_route.rb
mboneil10: Like I did below?

lib/tasks/route_import.rake
mboneil10: Ok I'll do local scopes then

app/models/bus_stops_route.rb
mboneil10: I hope I did it right

lib/tasks/route_import.rake
```diff
@@ -1,16 +1,58 @@
...
+      end
+    end
+
+    route_hash.each do |route, directions|
+      directions.each do |direction, variants|
+        sequences = []
+        stop_list = []
+
+        # sort the variants by the length of their stops, longest variant first
+        sorted_variants = variants.sort_by do |_variant, stops|
```
mboneil10: Omg this is great. Good thing ruby does stuff.

lib/tasks/route_import.rake
```diff
@@ -1,16 +1,58 @@
...
+            stop_hash.each do |hastus_id, rank|
+              sequence = rank.to_i
+              stop = BusStop.find_by hastus_id: hastus_id
+              next if stop_list.include? stop
+              stop_list << stop
+              if sequences.include? sequence
+                sequence = sequences.last + 1
+              end
+              sequences << sequence
+              BusStopsRoute.create!(
```
mboneil10: Ok, cool!

lib/tasks/stop_import.rake
```diff
@@ -10,11 +10,33 @@ STRING_COLUMN_ROW_NAMES = {
   mounting:           %w(stp_ud_mounting_pvta_pole         stp_ud_mounting_pole_other        stp_ud_mounting_structure  ),
   schedule_holder:    %w(stp_ud_schedule_holder_on_pole    stp_ud_schedule_holder_in_shelter                            ),
   shelter:            %w(stp_ud_shelter_pvta               stp_ud_shelter_other              stp_ud_shelter_building    ),
-  sidewalk:           %w(stp_ud_sidewalk_more_than_36in    stp_ud_sidewalk_less_than_36in    stp_ud_sidewalk_no_sidewalk),
-  sign:               %w(stp_ud_sign_flag_stop             stp_ud_sign_missing               stp_ud_sign_needs_attention),
+  sidewalk_width:           %w(stp_ud_sidewalk_more_than_36in    stp_ud_sidewalk_less_than_36in    stp_ud_sidewalk_no_sidewalk),
+  # sign:               %w(stp_ud_sign_flag_stop             stp_ud_sign_missing               stp_ud_sign_needs_attention),
+  # apparently sign's not needed :/
```
mboneil10: Yeah I guess we'll have to double check with everyone about this??

lib/tasks/stop_import.rake
```diff
@@ -39,7 +61,7 @@ namespace :bus_stops do
         attrs[db_column_name] = row[csv_column_name]
       end
       next if attrs[:name].blank?
-      BusStop.create! attrs
+      BusStop.find_or_create_by attrs
```
mboneil10: Sweet

mboneil10:
@sherson, is it still fine to rank according to other variant and place after the first variant?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 83.44
+    "covered_percent": 27.78
```
Anbranin: wat

app/models/bus_stops_route.rb
mboneil10: Did it right. Tested.

coverage/.last_run.json
mboneil10: Something with my local machine.

mboneil10: Or something else. Idfk. Looking into it.

PR #76, opened 2018-06-21
Fix pending bus stop scope

jforgue:
Fixes an issue where the pending scope also included completed bus stops.

screen shot 2018-06-21 at 10 22 06 am


jforgue:
Might as well slip in a custom validation message into this pull request.

PR #77, opened 2018-06-22
Order by sequence

Anbranin:
screen shot 2018-06-22 at 3 20 19 pm


app/views/bus_stops/by_sequence.haml
```diff
@@ -0,0 +1,13 @@
+.page-title Route #{@route.number} Bus Stops
+
+.container
+  - @collection.each do |direction, bsrs|
+    = direction + 'bound'
```
mboneil10: I called `BusStopsRoute.pluck(:direction).uniq` and got `["South", "North", "West", "East", "Downward", "Upward"]`. Maybe "downwardbound" looks strange, but it doesn't really matter.

app/views/bus_stops/index.haml
```diff
@@ -10,7 +10,11 @@ or
...
   = submit_tag 'Search'
 or
-= form_tag by_route_bus_stops_path, method: :get do
+= form_tag by_status_bus_stops_path, method: :get do
   = label_tag :number, 'Select a route'
   = select_tag :number, options_for_select(Route.pluck :number)
-  = submit_tag 'Search'
+  = submit_tag 'View stops by status'
+= form_tag by_sequence_bus_stops_path, method: :get do
+  = label_tag :number, 'Select a route'
```
mboneil10: Idk how I feel about the label that is repeated.

app/views/bus_stops/by_sequence.haml
dfaulken: Do we need the 'bound'?

mboneil10: No. Let's go without it.

PR #78, opened 2018-06-24
Additional Unit and System tests

mboneil10:
So... I've added some unit tests. And then started to test the edit stop page. There's an error happening in the context block of the new system spec, but I'm not worried about that being a problem with the app.


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 83.44
+    "covered_percent": 43.37
```
mboneil10: I still don't understand why when I push that the coverage is different. On my local machine, the coverage is up to 88.52%. Like seriously what. I'll investigate the coverage difference tomorrow, but the same thing happened in a previous branch. When @jforgue pulled that test branch of mine, ran rspec on his machine, and pushed, coverage went up. So it's definitely my problem.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 83.44
+    "covered_percent": 50.0
```
mboneil10: WTF UGH. I made a comment about this in a previous push.

mboneil10:
To be clear, the error I'm getting from trying to check 'Completed':
``` Failure/Error: check 'Completed'

     Selenium::WebDriver::Error::UnknownError:
       unknown error: Element <input class="triple-zoom" type="checkbox" value="1" name="bus_stop[completed]" id="bus_stop_completed"> is not clickable at point (270, 131). Other element would receive the click: <label for="bus_stop_has_power">...</label>
         (Session info: chrome=67.0.3396.87)
         (Driver info: chromedriver=2.35.528157 (4429ca2590d6988c0745c24c8858745aaaec01ef),platform=Mac OS X 10.12.5 x86_64)

     [Screenshot]: tmp/screenshots/failures_r_spec_example_groups_editing_a_bus_stop_as_a_user_with_a_bus_stop_that_has_been_completed_displays_who_updated_it_and_when_139.png
```

mboneil10:
Step 3 of 3 for #21 (Until the rewrite 😏 )

dfaulken:
Should folks be reviewing this @mboneil10?

dfaulken:
I think my next suggestion would be to remove the symbol loop from the factory, in favor of just listing the attributes outright.

spec/models/bus_stop_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+    end
+  end
+
+  describe 'completed scope' do
+    let!(:pending_stop) { create :bus_stop, :pending }
+    let!(:completed_stop) { create :bus_stop, :completed }
+    let!(:not_started_stop) { create :bus_stop }
+    it 'returns completed bus stops' do
+      expect(BusStop.all.completed).to include completed_stop
```
dfaulken: Does `BusStop.completed` not work?

spec/models/bus_stop_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    let(:stop) { create :bus_stop, :pending }
+    context 'bus stop completed attribute changed' do
+      context 'bus stop is completed' do
+        it 'assigns user to completed by' do
+          Timecop.freeze Date.today do
+            stop.update! accessible: true
+            stop.completed = true
+          end
+          stop.decide_if_completed_by user
+          expect(stop.completed_by).to be user
```
dfaulken: I'm not sure about `be` here. The two `User` instances will definitely be the same in the database, but I'm not sure about strict object-level equality.

spec/factories/bus_stops.rb
```diff
@@ -14,11 +14,11 @@
...
       created_at Date.yesterday
       updated_at Date.today
-        %i[bolt_on_base bus_pull_out_exists has_power solar_lighting
-          system_map_exists shelter_ada_compliance ada_landing_pad
-          state_road].each do |b|
-          send(b, 'true')
-          end
+      %i[bolt_on_base bus_pull_out_exists has_power solar_lighting
+        system_map_exists shelter_ada_compliance ada_landing_pad
+        state_road].each do |b|
```
dfaulken: Probably better to state them outright:

```ruby
bolt_on_base        true
bus_pull_out_exists true
```

It seems repetitive, but that's fine. I think the advantage is that it reads how a factory is supposed to read.

spec/factories/bus_stops.rb
```diff
@@ -14,11 +14,11 @@
...
+        send(b, 'true')
```
dfaulken: I advocate not having this line, but if we do, we don't want the *string* "true", right? we want `true`, the boolean literal?

spec/models/bus_stop_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'spec_helper'
+
+describe BusStop do
+  describe 'decide_if_completed_by' do
+    let(:user) { create :user }
+    let(:stop) { create :bus_stop, :pending }
+    context 'bus stop completed attribute changed' do
+      context 'bus stop is completed' do
+        it 'assigns user to completed by' do
+          Timecop.freeze Date.today do
```
dfaulken: Curious why we need to freeze to today.

spec/models/bus_stop_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      expect(BusStop.all.pending).to include pending_stop
+      expect(BusStop.all.pending).not_to include completed_stop
+      expect(BusStop.all.pending).not_to include not_started_stop
+    end
+  end
+
+  describe 'validations' do
+    context 'bus stop is not completed but assigned completed' do
+      it 'is not valid' do
+        invalid_stop = build :bus_stop, completed: true
```
dfaulken: Well, but if we change the factory to have more flushed-out attributes, then this fails without changing the test code.

Maybe just have one example of an attribute that's `nil` that shouldn't be?

spec/spec_helper.rb
```diff
@@ -48,3 +48,7 @@ def when_current_user_is(user)
 def format_time(time)
   time.strftime('%Y-%m-%d %H:%M %P')
 end
+
+def format_datetime(datetime)
+  datetime.strftime('%A, %B %e, %Y - %l:%M %P').squish
```
dfaulken: What's the `squish` for? I know what it does, but why do you need it?

spec/system/bus_stops/editing_bus_stops_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+    it 'updates the stop' do
+      expect(page).to have_selector 'p.notice',
+                                    text: 'Bus stop was updated.'
+    end
+    it 'redirects to the bus stops page' do
+      expect(page.current_url).to end_with bus_stops_path
+    end
+  end
+  #this context block is failing. TODO: figure out why
```
dfaulken: What kind of 'failing' is it?

spec/models/bus_stop_spec.rb
mboneil10: As opposed to what? Freezing to tomorrow? Or are you saying you want to know why I'm freezing.

spec/spec_helper.rb
mboneil10: I copied the method over.

spec/system/bus_stops/editing_bus_stops_spec.rb
mboneil10:
```
 Failure/Error: check 'Completed'

     Selenium::WebDriver::Error::UnknownError:
       unknown error: Element <input class="triple-zoom" type="checkbox" value="1" name="bus_stop[completed]" id="bus_stop_completed"> is not clickable at point (270, 131). Other element would receive the click: <label for="bus_stop_has_power">...</label>
         (Session info: chrome=67.0.3396.99)
         (Driver info: chromedriver=2.35.528157 (4429ca2590d6988c0745c24c8858745aaaec01ef),platform=Mac OS X 10.12.5 x86_64)
```

spec/models/bus_stop_spec.rb
dfaulken: Yes. why freeze at all, I'm not catching

mboneil10: Oh, in order to see that the attribute has changed.

dfaulken: Still confused. What happens if you don't freeze, I guess is what I'm asking.

spec/system/bus_stops/editing_bus_stops_spec.rb
dfaulken: Weird. It's trying to click on the checkbox, by clicking something else entirely?

Is this a byproduct of our CSS?

mboneil10: I think it is because of the CSS. I can't say for sure though.

spec/models/bus_stop_spec.rb
mboneil10: No difference 🐑

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 27.78
+    "covered_percent": 48.82
```
mboneil10: Idk what I did to fix this in other branches

spec/system/bus_stops/editing_bus_stops_spec.rb
jforgue: Have you tried removing the `triple-zoom` class? I'm curious as to why that would cause the issue.

spec/system/bus_stops/editing_bus_stops_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+  context 'with errors' do
+    before :each do
+      within 'table.edit-form' do
+        check 'Completed'
+        click_button 'Save stop'
+      end
+    end
+    it 'sends a helpful error message' do
+      expect(page).to have_selector 'p.error',
+                                    text: "Bench can't be blank"
```
jforgue: Won't `accessible` also be blank?

spec/factories/bus_stops.rb
```diff
@@ -5,21 +5,41 @@
...
+      shelter_pad_condition 'Great'
+      shelter_pad_material 'Asphalt'
+      shelter_type 'Modern half'
+      shared_sign_post 'Yes - FRTA'
+      need_work '5 - Immediate safety concern'
+      obstructions 'Yes - Bollard/Structure'
+      stop_sticker 'Sticker correct'
+      route_stickers 'No stickers'
+      bike_rack 'Bike locker'
+      real_time_information 'Yes -- Solar'
```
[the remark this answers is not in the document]
mboneil10: That's how the choice was listed in the app when I wrote this.

spec/factories/bus_stops.rb
```diff
@@ -5,21 +5,41 @@
...
-    # all attrs except for accessible
+      bolt_on_base true
+      bus_pull_out_exists true
+      has_power true
+      solar_lighting true
+      system_map_exists true
+      shelter_ada_compliance true
+      ada_landing_pad true
+      state_road true
+      # all attrs except for accessible
```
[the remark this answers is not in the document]
mboneil10: Not this attr because the trait is `pending`. If you would prefer that I fill in this field in and rename the trait to `completed`, fine. Then I could call `completed` and update an attr to nil in the `pending` trait.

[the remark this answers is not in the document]
mboneil10: I did it this way because this is what I prefer.

spec/models/bus_stop_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  describe 'completed scope' do
+    let!(:pending_stop) { create :bus_stop, :pending }
+    let!(:completed_stop) { create :bus_stop, :completed }
+    let!(:not_started_stop) { create :bus_stop }
+    it 'returns completed bus stops' do
+      expect(BusStop.completed).to include completed_stop
+      expect(BusStop.completed).not_to include pending_stop
+      expect(BusStop.completed).not_to include not_started_stop
```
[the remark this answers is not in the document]
mboneil10: Yes, `BusStop.completed` is `[completed_stop]`. But in this test, I don't particularly care that `BusStop.completed` is an array. What I'm testing for is that the elements inside of the array are correct. So, I can test the elements using `include` and `not_to include`

spec/factories/bus_stops.rb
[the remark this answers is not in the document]
mboneil10: I will change it though, since it isn't consistent.

spec/models/bus_stop_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+  end
+
+  describe 'not_started scope' do
+    let!(:pending_stop) { create :bus_stop, :pending }
+    let!(:completed_stop) { create :bus_stop, :completed }
+    let!(:not_started_stop) { create :bus_stop }
```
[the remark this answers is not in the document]
mboneil10: LOL whoops.

spec/system/bus_stops/editing_bus_stops_spec.rb
mboneil10: Yes I was just lazy.

mboneil10: A lot of other fields will be blank there.

spec/system/bus_stops/editing_bus_stops_spec.rb
mboneil10: Yeah, it fixes the problem when I remove the `triple-zoom` class

mboneil10:
What's preventing merging: testing the completed check box.

spec/system/bus_stops/editing_bus_stops_spec.rb
mboneil10: It's the `zoom` property. I'll use `transform: scale` from now on.

mboneil10:
Now that I've fixed the completed check box, I can't help but wonder. Is this _really_ what we want?
screen shot 2018-07-09 at 9 50 33 am

mboneil10:
I was thinking, could we do something like incidents does?
screen shot 2018-07-09 at 9 53 12 am

mboneil10:
I'll make this a new question issue, sorry out of scope of PR.

mboneil10:
Should I edit the index spec while I'm here?

mboneil10:
screen shot 2018-07-09 at 11 00 15 am

jforgue:
This breaks my heart a little bit.
42458144-4c4c9070-8367-11e8-822f-f362cd45375f

spec/models/bus_stop_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      expect(BusStop.pending).to include pending_stop
+      expect(BusStop.pending).not_to include completed_stop
+      expect(BusStop.pending).not_to include not_started_stop
+    end
+  end
+
+  describe 'validations' do
+    context 'bus stop is not completed but assigned completed' do
+      it 'is not valid' do
+        invalid_stop = build :bus_stop, completed: true, accessible: nil
```
jforgue: Isn't `accessible` already `nil` since it's not defined in the factory?

spec/system/bus_stops/editing_bus_stops_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+          within 'table.edit-form' do
+            click_button 'Save stop'
+          end
+        end
+        when_current_user_is user
+        visit edit_bus_stop_path(edit_stop.hastus_id)
+        expect(page).to have_content "Last updated by #{last_user.name} at #{format_datetime(1.day.ago)}"
+    end
+  end
+  # same issue as above context with check 'Completed'
```
jforgue: Has this been resolved yet?

spec/spec_helper.rb
jforgue: Seems like you could remove it then?

spec/factories/bus_stops.rb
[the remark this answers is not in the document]
dfaulken: So just to confirm, pending is everything except accessible (just as an example of an unfilled boolean value) and completed?

spec/models/bus_stop_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    context 'bus stop completed attribute changed' do
+      context 'bus stop is completed' do
+        it 'assigns user to completed by' do
+          stop.update! accessible: true
+          stop.completed = true
+          stop.decide_if_completed_by user
+          expect(stop.completed_by).to eql user
+        end
+      end
+      context 'bus stop is not completed' do
```
dfaulken: This looks like a potential false positive. I do see that the stop is pending, and so not completed, but I *also* see that the completed hasn't changed.

Since we haven't fulfilled *either* condition under which the method runs, logic dictates that this isn't an exhaustive test.

spec/models/bus_stop_spec.rb
dfaulken: Yes, but it's clearer to call it out here, in my opinion.

spec/spec_helper.rb
dfaulken: Or maybe not — but we should make sure we know why it's needed (or not)

spec/system/bus_stops/editing_bus_stops_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    when_current_user_is user
+    visit edit_bus_stop_path(edit_stop.hastus_id)
+  end
+  context 'with no changes' do
+    before :each do
+      within 'table.edit-form' do
+        click_button 'Save stop'
+      end
+    end
+    it 'updates the stop' do
```
dfaulken: Strictly speaking this doesn't do what it says on the tin. A message saying that the stop wasn't updated isn't proof positive that the stop was updated.

We could be huge liars.

spec/system/bus_stops/editing_bus_stops_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+        Timecop.freeze 1.day.ago do
+          when_current_user_is last_user
+          visit edit_bus_stop_path(edit_stop.hastus_id)
+          within 'table.edit-form' do
+            click_button 'Save stop'
+          end
+        end
+        when_current_user_is user
+        visit edit_bus_stop_path(edit_stop.hastus_id)
+        expect(page).to have_content "Last updated by #{last_user.name} at #{format_datetime(1.day.ago)}"
```
dfaulken: Very nice

spec/factories/bus_stops.rb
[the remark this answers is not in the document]
mboneil10: yes

spec/system/bus_stops/editing_bus_stops_spec.rb
mboneil10: OH yes it has.

spec/spec_helper.rb
mboneil10: ok

spec/models/bus_stop_spec.rb
mboneil10: :( ok

spec/system/bus_stops/editing_bus_stops_spec.rb
mboneil10: It doesn't always pass. 👍

mboneil10: This only happens sometimes.

mboneil10: Is it worth it to `timecop`?

spec/models/bus_stop_spec.rb
mboneil10: I've updated it?

mboneil10:
@jforgue, can you finish the styling for this? You don't have to.

spec/factories/bus_stops.rb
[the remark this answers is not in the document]
dfaulken: Okay, in that case I might recommend doing something other than accessible as an example - something that seems as arbitrary as possible of an exclusion to make.

One could misinterpret this as a point of logic - for instance that a stop is required to be marked as accessible before it can be marked as completed.

spec/models/bus_stop_spec.rb
dfaulken: Yes, looks correct now.

spec/factories/bus_stops.rb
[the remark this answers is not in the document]
mboneil10: Ah. I'm not sure how to do that without logic in the factory.

mboneil10:
@dfaulken I agree with you. But what would you suggest I do? I don't want to put logic in the factory. (I thought I wrote this comment but it somehow got lost).

dfaulken:
Just pick a different, random attribute. One that isn't a boolean.

mboneil10:
OH

mboneil10:
screen shot 2018-07-16 at 11 47 13 am

dfaulken:
@mboneil10, sorry, realized I might've misread which comment of mine your question was referring to? Were you referring to my main comment accompanying the PR review, or the comment on the specific choice of `accessible` in the factory?

mboneil10:
@dfaulken I was referring to the choice of `accessible` in the factory. I actually hadn't seen the other one.

mboneil10:
But as for the second comment, I'm not entirely sure how we could do that. :(

dfaulken:
> But as for the second comment, I'm not entirely sure how we could do that. :( Could we do something like a factory for fields?

I don't know, offhand. We would need to think more about the best way to accomplish this.

spec/system/bus_stops/managing_bus_stops_spec.rb
```diff
@@ -38,7 +38,9 @@
 describe 'viewing outdated' do
   let(:admin) { create :user, :admin }
   let!(:present_stop) { create :bus_stop }
-  let!(:date) { Date.today }
+  # default date for outdated is from a month ago
+  let!(:date) { Date.today - 1.month }
+  let(:picked_date) { date.change(day: 28) }
```
dfaulken: I don't see that we use this anywhere.

mboneil10: Do you mean [`picked_date`](https://github.com/umts/stop-project/pull/78/files/a90992220528668858c9c24164297dfdff6292ed#diff-feabba5eae22c49551b0e820f7aae3bdR82) or [`date`](https://github.com/umts/stop-project/pull/78/files/a90992220528668858c9c24164297dfdff6292ed#diff-feabba5eae22c49551b0e820f7aae3bdL42)? Both are used, but maybe they don't need to. Would you suggest that I ditch `date` and use `Date.today - 1.month` wherever it's needed?

dfaulken: No, I just misread and didn't see where `picked_date` is used. Disregard.

jforgue:
@mboneil10 Sorry, I missed the email and just saw your comment today.

spec/system/bus_stops/managing_bus_stops_spec.rb
mboneil10: Okie dokie

PR #80, opened 2018-06-25
Revert "Revert "Route variants""

Anbranin:
Reverts umts/stop-project#79


app/controllers/bus_stops_controller.rb
```diff
@@ -8,7 +8,18 @@ def autocomplete
     render json: stops.pluck(:name).sort
   end
 
-  def by_route
+  def by_sequence
+    @route = Route.find_by number: params.require(:number)
+    if @route.present?
+      @collection = @route.bus_stops_routes.group_by(&:direction).each do |_dir, bsrs|
+        bsrs.sort_by(&:sequence)
```
dfaulken: Are we confident that `each` and `sort_by` is actually doing anything?

I would expect that we would need either `each` and `sort_by!` (procedural notation) or `map` and `sort_by` (functional notation).

Is there a test for this?

app/models/bus_stops_route.rb
```diff
@@ -0,0 +1,9 @@
+class BusStopsRoute < ApplicationRecord
+  belongs_to :bus_stop
+  belongs_to :route
+
+  validates :sequence, presence: true, numericality: { greater_than_or_equal_to: 1 }
+  validates :direction, presence: true
+  validates_uniqueness_of :sequence, scope: [ :route, :direction ]
```
dfaulken: Shouldn't have spaces inside of square brackets.

app/views/bus_stops/index.haml
```diff
@@ -10,7 +10,11 @@ or
...
 or
-= form_tag by_route_bus_stops_path, method: :get do
+= form_tag by_status_bus_stops_path, method: :get do
   = label_tag :number, 'Select a route'
   = select_tag :number, options_for_select(Route.pluck :number)
-  = submit_tag 'Search'
+  = submit_tag 'View stops by status'
+= form_tag by_sequence_bus_stops_path, method: :get do
+  = label_tag :number, 'Select a route'
+  = select_tag :number, options_for_select(Route.pluck :number)
```
dfaulken: Omitting inner parentheses is generally something we've avoided as a point of style. But since `Route.pluck(:number)` is a piece of data, better to fetch it in the controller and sidestep the style issue altogether.

config/routes.rb
```diff
@@ -13,12 +13,13 @@
...
       post :autocomplete
-      get  :by_route
+      get  :by_sequence
+      get  :by_status
       post :id_search
       get  :manage
       post :name_search
       get  :outdated
-      get :field_guide
+      get  :field_guide
```
dfaulken: This is out of alphabetical order.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 83.44
+    "covered_percent": 27.78
```
dfaulken: We really need to nail down what's up with your test environment @mboneil10.

db/seeds.rb
```diff
@@ -41,14 +43,31 @@
...
+      sequence = 0
+    else
+      sequence = stop_names.length
+    end
+    stop_names.each do |stop_name|
+      if direction == 'North'
+        sequence += 1
+      else
+        sequence -= 1
+      end
```
dfaulken: That's an interesting algorithm. I think, like in the rake task, it would be simpler to construct the lists of stops first, then give them sequence numbers according to where they are in the list (`each.with_index(1)`, etc.).

This is *clever*, but not particularly intuitive, in my view.

config/routes.rb
jforgue: Nit pick, but we don't need this space

db/seeds.rb
```diff
@@ -41,14 +43,31 @@
...
+      # Anytime in the last two months
+      Timecop.freeze rand(86_400).minutes.ago do
```
dfaulken: Didn't I redo this to actually read two months? Or is that somewhere not merged?

coverage/.last_run.json
jforgue: Looks like we didn't ruin `rspec`.

lib/tasks/stop_import.rake
```diff
@@ -10,11 +10,33 @@ STRING_COLUMN_ROW_NAMES = {
   mounting:           %w(stp_ud_mounting_pvta_pole         stp_ud_mounting_pole_other        stp_ud_mounting_structure  ),
   schedule_holder:    %w(stp_ud_schedule_holder_on_pole    stp_ud_schedule_holder_in_shelter                            ),
   shelter:            %w(stp_ud_shelter_pvta               stp_ud_shelter_other              stp_ud_shelter_building    ),
-  sidewalk:           %w(stp_ud_sidewalk_more_than_36in    stp_ud_sidewalk_less_than_36in    stp_ud_sidewalk_no_sidewalk),
-  sign:               %w(stp_ud_sign_flag_stop             stp_ud_sign_missing               stp_ud_sign_needs_attention),
+  sidewalk_width:           %w(stp_ud_sidewalk_more_than_36in    stp_ud_sidewalk_less_than_36in    stp_ud_sidewalk_no_sidewalk),
+  # sign:               %w(stp_ud_sign_flag_stop             stp_ud_sign_missing               stp_ud_sign_needs_attention),
+  # apparently sign's not needed :/
```
dfaulken: So then we can just remove, right? Rather than comment?

Bearing in mind that this will all be replaced anyway.

lib/tasks/stop_import.rake
```diff
@@ -10,11 +10,33 @@ STRING_COLUMN_ROW_NAMES = {
...
+  schedule_holder: ['On pole', 'In shelter'],
+  shelter: %w[PVTA Other Building],
+  shelter_condition: %w[Great Good Fair Poor],
+  shelter_pad_condition: %w[Great Good Fair Poor],
+  shelter_pad_material: %w[Asphalt Concrete Other],
+  shelter_type: %w[Modern Modern\ half Victorian Dome Wooden Extra\ large Other],
+  sidewalk_width: ['More than 36"', 'Less than 36"', 'None'],
+  # sign: ['Flag stop', 'Missing sign', 'Needs attention'],
+  # sign_type: ['Axehead (2014+)', 'Rectangle (<2014)', 'MGM + Axhead (2018+)'],
+  # sign type is not in hastus yet
```
dfaulken: But it doesn't matter. If it's in the field guide, we need to be collecting the data. We have quite a while to make HASTUS match the field guide. This app needs to match it now.

spec/factories/bus_stops_routes.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :bus_stops_route do
+    sequence :sequence
```
dfaulken: Nice.

db/seeds.rb
```diff
@@ -41,14 +43,31 @@
...
-    Timecop.freeze rand(86_400).minutes.ago do
-      stop = BusStop.find_or_initialize_by name: stop_name
-      stop.hastus_id = hastus_ids.fetch(stop_name)
-      stop.routes << routes.fetch(route_number)
-      stop.save!
+  directions.each do |direction|
+    if direction == 'North'
+      sequence = 0
+    else
+      sequence = stop_names.length
```
jforgue: Can someone please explain why `stop_name.length`? is this arbitrary?

coverage/.last_run.json
mboneil10: I did, problem with me. Idfk what's happening with it.

config/routes.rb
jforgue: Oh I see this was too far left, nvm my bad.

app/models/bus_stops_route.rb
mboneil10: Whoops, got it.

coverage/.last_run.json
mboneil10: Yes.

db/seeds.rb
mboneil10: That's in the `rewrite` branch.

lib/tasks/stop_import.rake
mboneil10: True. No need for umaps rails mistakes in here.

db/seeds.rb
mboneil10: Dammit. Okay.

db/seeds.rb
dfaulken: The algorithm counts down from the length of the array for southbound stops.

app/views/bus_stops/index.haml
```diff
@@ -10,7 +10,11 @@ or
...
+  = submit_tag 'View stops by sequence'
```
jforgue: There are two submit tags on this page doing two different things. Seems both confusing and possibly problematic.

db/seeds.rb
mboneil10: Edit: not a bus driver. Don't come at me for having stops that aren't in the same route, guys :)

app/controllers/bus_stops_controller.rb
mboneil10: No test for this.

mboneil10: This is what I'll be working on.

db/seeds.rb
mboneil10: Sorry dave didn't see your comment

dfaulken:
> Looks like @dfaulken got all the functional stuff.

To clarify my initial review comment, I want to look at everything except the rake task first (that is, the stuff that operates on the data that has been imported).

Then I want to dive into the rake task (the stuff that imports the data that everything else operates on) as a separate point of review.

db/seeds.rb
jforgue: I see, so `stop_name` is an array containing the stops? Not the actual name of the stop?

app/views/bus_stops/index.haml
mboneil10: I agree.

spec/system/index_spec.rb
```diff
@@ -83,22 +83,44 @@
...
+  let!(:bus_stops_route) { create :bus_stops_route, bus_stop: bus_stop, route: route }
   before :each do
     when_current_user_is user
     visit root_url
   end
   context 'route from the dropdown' do
-    it 'redirects to bus stops by route page' do
+    it 'redirects to bus stops by sequence' do
       within 'form', text: 'Select a route' do
         select route.number, from: 'Select a route'
```
jforgue: This is ambiguous, view needs to be corrected.

mboneil10: :(

mboneil10: @jforgue is fixing the view while I'm fixing the test

db/seeds.rb
mboneil10: 'stop_names' is an array containing the stop names. If you look at the `each` loop, it's initially using `stops` (defined [here](https://github.com/umts/stop-project/pull/80/files/e4d0ddb5c023e8c21c1361a6dece852737165ce9#diff-f7a25bc30a409fc77050703fc048ca8cL21)).

app/views/bus_stops/by_sequence.haml
```diff
@@ -0,0 +1,13 @@
+.page-title Route #{@route.number} Bus Stops
+
+.container
```
dfaulken: Confused by why we need a container at this level. Isn't there one in the layout?

app/views/bus_stops/by_sequence.haml
```diff
@@ -0,0 +1,13 @@
...
+  - @collection.each do |direction, bsrs|
+    = direction
+    - bsrs.each do |bus_stop_route|
+      .row
+        .col-sm-1=bus_stop_route.sequence
```
dfaulken: Why is it that we want to show the sequence number here?

dfaulken: Also, here and throughout, we have usually stuck to the convention of leaving a space after `-` or `=` in HAML.

app/views/bus_stops/by_status.haml
```diff
@@ -6,7 +6,7 @@
...
     %ul
-    - stops.each do |stop|
-      %li
-        = link_to edit_bus_stop_path(stop.hastus_id) do
-          - stop.name
\ No newline at end of file
+      - stops.each do |stop|
+        %li
+          = link_to edit_bus_stop_path(stop.hastus_id) do
+            - stop.name
```
dfaulken: This... doesn't seem right. Surely we want to *display* the name of the stop, rather than just evaluating it as Ruby code.

db/seeds.rb
```diff
@@ -41,14 +43,31 @@
...
+        sequence -= 1
+      end
+      # Anytime in the last two months
+      Timecop.freeze rand(86_400).minutes.ago do
+        stop = BusStop.find_or_initialize_by name: stop_name
+        stop.hastus_id = hastus_ids.fetch(stop_name)
+        stop.save!
+        
+        route = routes.fetch(route_number)
+        BusStopsRoute.create route: route, sequence: sequence, bus_stop: stop, direction: direction
```
dfaulken: We probably want `create!` in the seeds file, so that we know if something isn't working.

spec/factories/bus_stops.rb
```diff
@@ -4,5 +4,27 @@
   factory :bus_stop do
     name 'stop'
     sequence :hastus_id
+    trait :pending do
+      %i[name bench curb_cut lighting mounting mounting_direction
+        schedule_holder shelter sidewalk_width trash mounting_clearance
+        sign_type shelter_condition shelter_pad_condition shelter_pad_material
+        shelter_type shared_sign_post garage_responsible need_work obstructions
+        stop_sticker route_stickers bike_rack real_time_information].each do |s|
+        send(s, 'not blank')
```
dfaulken: We should definitely be providing valid values in the factory. These are not valid, correct?

spec/factories/bus_stops.rb
```diff
@@ -4,5 +4,27 @@
...
+        end
+      created_at Date.yesterday
+      updated_at Date.today
+        %i[bolt_on_base bus_pull_out_exists has_power solar_lighting
+          system_map_exists shelter_ada_compliance ada_landing_pad
+          state_road].each do |b|
+          send(b, 'true')
```
dfaulken: These are booleans, so just `true`, right? Not a string.

In general, I think it would be better to set the attributes outright rather than in a loop - again, the least amount of logic you can write in a test environment, the better.

spec/factories/bus_stops.rb
```diff
@@ -4,5 +4,27 @@
...
+          end
+    # all attrs except for accessible
+    end
+    trait :completed do
+      pending
```
dfaulken: Didn't know you can compose traits like that. Neat!

spec/system/bus_stops/viewing_bus_stops.rb
```diff
@@ -0,0 +1,55 @@
...
+  let!(:stop_3) { create :bus_stop, :completed }
+  let!(:bsr1) { create :bus_stops_route, route: route, bus_stop: stop_1 }
+  let!(:bsr2) { create :bus_stops_route, route: route, bus_stop: stop_2 }
+  let!(:bsr3) { create :bus_stops_route, route: route, bus_stop: stop_3 }
+  before :each do
+    when_current_user_is user
+    visit by_sequence_bus_stops_path(number: route.number)
+  end
+  context 'with stops' do
+    it 'displays according to sequence' do
```
dfaulken: Well, okay, but this isn't a test that makes sure that things are in the right order.

spec/factories/bus_stops.rb
mboneil10: :) Thank you!

spec/system/bus_stops/viewing_bus_stops.rb
mboneil10: Oh yeah that's fixed in a newer version

PR #81, opened 2018-06-25
View changes

jforgue:


app/views/bus_stops/by_sequence.haml
```diff
@@ -1,13 +1,16 @@
 .page-title Route #{@route.number} Bus Stops
 
-.container
+%div
+  = link_to 'View by status', by_status_bus_stops_path(number: @route.number)
+- if @collection.first.first.nil?
```
dfaulken: Could you explain `.first.first` to me?

app/views/bus_stops/by_status.haml
```diff
@@ -1,5 +1,7 @@
 .page-title Route #{@route.number} Bus Stops
 
+%div
+  = link_to 'View by sequence', by_sequence_bus_stops_path(number: @route.number)
```
dfaulken: Let's come up with a better name. 'View by route order', maybe?

app/views/bus_stops/by_sequence.haml
```diff
@@ -1,13 +1,16 @@
 .page-title Route #{@route.number} Bus Stops
 
-.container
+%div
```
dfaulken: Why do we want this link in a div?

jforgue: It was the simplest way to keep it out of line with the next text, other than a few `br` tags .

app/views/bus_stops/by_sequence.haml
mboneil10: He wanted to check that `stops` wasn't empty. Probably better to put `@stops = @route.bus_stops` in the controller and rewrite the if?

mboneil10: Hm. I talked to @jforgue and it seems like I'm wrong. I'll let him explain.

PR #87, opened 2018-06-26
Update fields to match latest field guide

jforgue:


sherson:
Before this is deployed, we need to:

1. Coordinate a time with PVPC and UMTS Service to do this with impacting their use of the app (probably after 5 pm and before 9 am). I can take care of this.
2. Ensure that the existing data is saved. 
3. Confirm that we know what existing data in which fields will be affected. 
4. Determine who is going to deploy.
5. Come up with a plan to convert the data afterwards, and confirm that it's correct:

trash:
* `PVTA`, `Municipal`, `Other` = yes
* `None` = no

has_power:
* yes = `Yes - Stub up` (we might also have a list of stops that we need to set to `Yes - Outlet`)
* no = `No`

shared_sign_post:
* `Yes - Traffic sign`, `Yes - FRTA`, `Yes - Other`, `Sign not on pole` = yes
* `No` = no

system_map_exists:
* yes = ? (I will contact PVPC about this - it'll be`New map` or `Old map`) 
* no = `No map`

After it's deployed, we need to:

1. Convert the data.
2. Confirm that it's correct.
3. Inform PVPC and UMTS Service that it's good to go.

If I missed anything, anyone feel free to chime in.

@jforgue can you double-check that what I wrote above looks correct?

After that, @mboneil10 is going to go through this list, create and assign issues.

mboneil10:
Created corresponding issues with labels `before next deploy` and `after next deploy`.

jforgue:
Should not to be merged until #102 is merged and export executed.

app/views/bus_stops/_form.haml
```diff
@@ -38,7 +38,7 @@
           %tr
             %td= f.label name
             - if options.kind_of?(Array)
-              %td= f.select name, options, include_blank: true
+              %td= f.select name, options |= [stop.send(name)], include_blank: true
```
dfaulken: Interesting. Didn't know `|=` was a thing.

That being said, how feasible would it be to construct this in the controller? Or perhaps a model method?

jforgue: Sounds very doable. However let me ask this. I assuming this will be changed in the rewrite so, would the method become obsolete or is that better since the we will always need to display the old data?

dfaulken: Good question. I guess it's a question of how we think it'll be implemented in the refactor, and whether that's something we can build towards now.

mboneil10: I find logic in the view to be risky. If we kept it like this for now, I'd recommend that we rewrite it in the refactor anyway. I know @jforgue is working on this currently, and I'm glad.

jforgue: I've been trying to abstract this out, but it doesn't seem feasible to construct this in the controller or model. It seems the only way to not have logic in the view is by changing the options in the model before I pass them to the view. Which seems more texting and complex then a simple `|=`.

dfaulken: What about something like this in the controller:

```ruby
@fields = BusStop::SUPER_HASH # we shouldn't be using model constants in the view anyway
@fields.each_pair do |category, fields|
  fields.each_pair do |field, options|
    if options.is_a?(Array) && options.exclude?(@stop.send(field))
      @fields[category][field] << @stop.send(field)
    end
  end
end
```

Issue #98, opened 2018-06-26
Exclude non-public routes and stops from Hastus import


sherson:
https://github.com/umts/stop-project/tree/master/contrib

#97 (only loosely related)

mboneil10:
uhhh

dfaulken:
Relevant Hastus columns are `stp_is_public` and `rte_for_public_info`, I believe.

dfaulken:
@mboneil10 done correct? The PRs were merged, but looks like that might not have closed the issue.

mboneil10:
Yes, thanks for going through these.

PR #99, opened 2018-06-26
Exclude non-public routes from route import OIG

dfaulken:
Addresses #98.

This is just a proposed code change. I have not tested this OIG.

### DIFF
--- a/contrib/stops_with_routes.id
+++ b/contrib/stops_with_routes.id
@@ -13,7 +13,7 @@ file stops_with_routes {
     item 5 { value '"stop_variant_rank"' }
   }
   foreach trip {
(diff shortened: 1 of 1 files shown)


contrib/stops_with_routes.id
```diff
@@ -13,7 +13,7 @@ file stops_with_routes {
     item 5 { value '"stop_variant_rank"' }
   }
   foreach trip {
-    condition 'trp_via_variant <> NULL'
+    condition 'trp_via_variant <> NULL AND Object(route, trp_route).rte_for_public_info'
```
mboneil10: What's a trip route again?

dfaulken: The route the trip runs on.

PR #100, opened 2018-06-26
Exclude non-public stops

dfaulken:
Addresses #98.

This is just a proposed code change. I have not tested this OIG.


contrib/stops_for_stop_project.id
```diff
@@ -39,6 +39,7 @@ file stops_details_with_routes
       }
       
    foreach stop
+      condition 'stp_is_public'
```
mboneil10: And this is just a field in hastus?

Issue #92, opened 2018-06-26
Plan to Convert Saved Data


mboneil10:
Make a plan for this and confirm that it's correct. Related to https://github.com/umts/stop-project/issues/89

mboneil10:
Read the saved file from #89 and restore those values. Probably through a rake task.

mboneil10:
3 of 3 for existing data

Issue #89, opened 2018-06-26
Ensure Existing Data is saved


mboneil10:
@dfaulken's idea is to save it in a CSV or JSON file.

mboneil10:
part 1 of 3 for existing data

Issue #90, opened 2018-06-26
Confirmation of Existing Data that will be Lost


mboneil10:
Related to https://github.com/umts/stop-project/issues/89. We want to be sure that we are saving the right data and not forgetting anything.

mboneil10:
`trash:`

PVTA, Municipal, Other = yes
None = no
`has_power:`

yes = Yes - Stub up (we might also have a list of stops that we need to set to Yes - Outlet)
no = No
`shared_sign_post:`

Yes - Traffic sign, Yes - FRTA, Yes - Other, Sign not on pole = yes
No = no
`system_map_exists:`

yes = ? (I will contact PVPC about this - it'll beNew map or Old map)
no = No map

mboneil10:
Oh yeah, I was confused at the time because I was still under the impression that we were going to add new columns and migrate away disused columns whenever the data was converted. 

This is the original comment that @sherson had for [plans,](https://github.com/umts/stop-project/pull/87) not sure if you've seen it since the PR is closed. That's where I'm getting the labels for `before` and `after deploy`, along with all of these title. We can all talk on the channel about the order of this plan if you want.

PR #102, opened 2018-06-28
Stop data export and import tasks

jforgue:


lib/tasks/stop_data_convert_and_import.rake
```diff
@@ -0,0 +1,18 @@
...
+  task :import_data, [:csv_file] => :environment do |_, args|
+    csv = CSV.parse(File.read(args[:csv_file]), headers: true)
+    csv.each do |row|
+      hash = row.to_hash
+      hash['trash'] = (hash['trash'] != 'None')
+      hash['has_power'] = (hash['has_power'] == 'true' ? 'Yes - Stub up' : 'No')
+      hash['shared_sign_post'] = (hash['shared_sign_post'] == 'Yes - FRTA')
+      hash['shared_sign_post_frta'] = hash.delete('shared_sign_post')
+      hash['system_map_exists'] = (hash['system_map_exists'] == 'true' ? 'Old map' : 'No map')
+      stop = BusStop.find_by_name(hash['name'])
```
dfaulken: We shouldn't use name here. We've established that bus stops can have the same name.

lib/tasks/stop_data_export.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'csv'
+namespace :bus_stops do
+  task export: :environment do
+    CSV.open('old_stop_data.csv', 'w', write_headers: true,
```
dfaulken: What does `write_headers: true` do?

lib/tasks/stop_data_export.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'csv'
```
dfaulken: Newline after require statements please

lib/tasks/stop_data_export.rake
dfaulken: Also, why allow for specifying the CSV name in one task, but hardcode it in the other?

lib/tasks/stop_data_convert_and_import.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'csv'
+namespace :bus_stops do
+  task :import_data, [:csv_file] => :environment do |_, args|
+    csv = CSV.parse(File.read(args[:csv_file]), headers: true)
+    csv.each do |row|
+      hash = row.to_hash
```
dfaulken: I think it's a good idea to separate the data we're reading from the data we're writing - that is, to use two separate data structures. One that we don't touch, which is our input data, and one which starts out empty and which we do modify, which becomes the attributes.

jforgue: I understand what you're saying but could you please explain why this is necessary? If we're not saving over the CSV why does it matter how we modify it in the task?

dfaulken: Clarified in person that it's more about the code readability than its function necessarily.

lib/tasks/stop_data_convert_and_import.rake
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+namespace :bus_stops do
+  task import_old_data: :environment do
+    csv = CSV.parse(File.read('old_stop_data.csv'), headers: true)
```
dfaulken: I think there's a method that rolls up `CSV.parse` and `File.read` into one. `CSV.read` maybe?

lib/tasks/stop_data_convert_and_import.rake
```diff
@@ -0,0 +1,29 @@
...
+    csv.each do |row|
+      import_data = row.to_hash
```
dfaulken: Curious why we need to `to_hash` the `CSV::Row`?

lib/tasks/stop_data_convert_and_import.rake
```diff
@@ -0,0 +1,29 @@
...
+                             'system_map_exists' =>
+                                 (import_data['system_map_exists'] == 'true' ? 'Old map' : 'No map'),
+                             'trash' => (import_data['trash'] != 'None') }
+      conversion_methods.each do |attribute, conversion|
+        if attribute == 'shared_sign_post_frta'
+          new_attributes[attribute] = conversion unless import_data['shared_sign_post'].nil?
+        else
+          new_attributes[attribute] = conversion unless import_data[attribute].nil?
+        end
+      end
```
dfaulken: I still think this is overengineering. IMO it would be easier to read to just write:

```ruby
new_attributes['has_power'] = case import_data['has_power']
                              when true then 'Yes - Stub up'
                              when false then 'No'
                              end
```

etc. It's a little less DRY, but there's *no* special logic at all built in. You can read it once and understand exactly what's going on.

Especially important since this is something we will only have to look back it if something else has gone wrong. Those are the moments when clever code is your worst enemy.

lib/tasks/stop_data_export.rake
```diff
@@ -0,0 +1,19 @@
...
+
+namespace :bus_stops do
+  task export_old_data: :environment do
+    CSV.open('old_stop_data.csv', 'w') do |csv|
+      headers = %w[id has_power shared_sign_post system_map_exists trash]
```
dfaulken: So is *this* a case where we could use `write_headers: true` and then shovel hashes of attributes to the CSV rather than arrays?

jforgue: That sounds, right but I'm having trouble figuring out what this would look like. Can you with an example? 
I tried something like this but only to get a no method error.
`csv << { id: stop.id, has_power: stop.has_power,`
                ` shared_sign_post: stop.shared_sign_post,`
                 `system_map_exists: stop.system_map_exists,`
                 `trash: stop.trash }`

dfaulken: I'm going to change my suggestion to be that we specify the headers in the argument to `open`, and use `write_headers: true`, now that I've actually read the docs. https://ruby-doc.org/stdlib-2.0.0/libdoc/csv/rdoc/CSV.html#method-c-new

But it seems you're correct, it'll have to be an array.

lib/tasks/stop_data_convert_and_import.rake
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+namespace :bus_stops do
+  task import_old_data: :environment do
+    csv = CSV.read('old_stop_data.csv', headers: true)
+    csv.each do |row|
+      new_attributes = {}
```
mboneil10: Why is this called `new_attributes`?

jforgue: I suppose `new_values` might make more sense, although the attribute types change.

mboneil10: I typically stay away from key words like `new`, but that's a personal preference. I know you didn't know how to do a rake task before, thanks for sticking with it! Good job

mboneil10:
Closes #89 and #90 (exports existing data to CSV). Also closes #92 (import saved data).

lib/tasks/stop_data_convert_and_import.rake
dfaulken: It's not a keyword here, though.

lib/tasks/stop_data_convert_and_import.rake
```diff
@@ -0,0 +1,32 @@
...
+      new_attributes['system_map_exists'] = case row['system_map_exists']
+                                            when 'true' then 'Old map'
+                                            when 'false' then 'No map'
+                                            end
+      new_attributes['trash'] = case row['trash']
+                                when nil then nil
+                                when 'None' then false
+                                else true
+                                end
+      stop = BusStop.find_by_id(row['id'])
```
dfaulken: We should avoid the `find_by_xxx` in Rails >3. I would prefer `find_by id: row['id']` here, though bear in mind that's the same as `find row['id']`.

lib/tasks/stop_data_convert_and_import.rake
mboneil10: Ah ok no worries then

PR #103, opened 2018-06-29
BusStopRoutes import and Test

mboneil10:


spec/models/bus_stops_route_spec.rb
```diff
@@ -35,4 +35,74 @@
...
+                          'o_North' => other_variant } }
+          BusStopsRoute.import(stop_hash)
+          expect(stop_hash[route_dir]).to eql sequenced_stops
+        end
+      end
+    end
+    context 'first stop is not in longest variant' do
+      context 'one stop is in other variant' do
+        it 'sequences stops correctly' do
+          sequenced_stops = %w[A B C F D E]
```
mboneil10: I know this pattern repeats itself. Should I -- and if so, how do I -- abstract it out?

mboneil10:
@dfaulken would you mind checking my tests? And @jforgue, the correctness of the algorithm?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 27.78
+    "covered_percent": 83.41
```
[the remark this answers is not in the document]
mboneil10: finally.

spec/models/bus_stops_route_spec.rb
mboneil10: Ah, okay. I was having trouble with the `doesnt_matter` parts. Hooray!

mboneil10: Oh, got it

mboneil10:
@jforgue ready for review

spec/models/bus_stops_route_spec.rb
```diff
@@ -35,4 +35,57 @@
...
+        let(:other_variant) { %w[F G H D E] }
+        it 'sequences stops correctly' do
+          resequence!
+          expect(@stop_hash[:route_dir]).to eql %w[A B C F G H D E]
+        end
+      end
+    end
+    context 'no stop is in the longest variant' do
+      let(:other_variant) { %w[F G H I] }
+      it 'sequences stops correctly' do
```
dfaulken: More descriptive descriptions would be nice throughout.

spec/models/bus_stops_route_spec.rb
```diff
@@ -35,4 +35,57 @@
       end
     end
   end
+  describe 'import' do
+    let(:main_variant) { %w[A B C D E] }
+    before :each do
+        @stop_hash = { route_dir: { key_1: main_variant, key_2: other_variant } }
+    end
+    let(:resequence!) { BusStopsRoute.import(@stop_hash) }
+    let(:other_variant) { [] }
```
dfaulken: I don't think it's possible for the other variant to be truly empty, right?

lib/tasks/route_import.rake
```diff
@@ -6,52 +6,29 @@ require 'csv'
...
-        route_hash[route] ||= {}
-        route_hash[route][direction] ||= {}
-        route_hash[route][direction][variant] ||= []
-        route_hash[route][direction][variant] << { stop_id => sequence }
-      end
+      route = row['rte_identifier']
+      stop = row['stp_identifier']
+      dir = row['direction']
+      variant = row['variant']
+      sequence = row['stop_variant_rank'].to_i
```
dfaulken: Nitpick, but I might define the variables in the order you use them (or perhaps in the order they occur in the file).

app/models/bus_stops_route.rb
```diff
@@ -9,4 +9,45 @@ class BusStopsRoute < ApplicationRecord
   validates :direction, presence: true
   validates_uniqueness_of :sequence, scope: %i[route direction]
   validates_uniqueness_of :bus_stop, scope: %i[route direction]
+
+  def self.import(stop_hash)
```
dfaulken: It might be nice to outline the algorithm in a block comment before the method body.

app/models/bus_stops_route.rb
```diff
@@ -9,4 +9,45 @@ class BusStopsRoute < ApplicationRecord
...
+    stop_hash.each_pair do |route_dir, direction_variants|
+      variants = direction_variants.values
```
dfaulken: Might be worth explaining why the keys are there at all, if they're not used for this method.

app/models/bus_stops_route.rb
```diff
@@ -9,4 +9,45 @@ class BusStopsRoute < ApplicationRecord
...
+      longest_variant = variants.max_by(&:length)
+      other_variants = variants - longest_variant
+      other_variants.each do |other_variant|
+        other_variant.each.with_index do |stop, sequence|
+          unless longest_variant.include? stop
+            # if sequence is zero, then it is the first stop. We can't rely on
+            # its previous stop like we do in the corresponding else statement.
+            if sequence == 0
```
dfaulken: In case you would prefer, you could also write `.zero?`

But perhaps in this case `if stop == other_variant.first` might be the *most* communicative?

app/models/bus_stops_route.rb
```diff
@@ -9,4 +9,45 @@ class BusStopsRoute < ApplicationRecord
...
+              else
+                # add the stop onto the end of the longest variant
+                longest_variant << stop
+              end
+            else
+              # the previous stop will already be in the longest variant
+              previous_stop = other_variant[sequence - 1]
+              previous_stop_index_in_longest_variant = longest_variant.index(previous_stop)
+              # Array#insert inserts before the given index.
+              longest_variant.insert(previous_stop_index_in_longest_variant + 1, stop)
```
dfaulken: Feel free to disagree on my verbose variable names.

spec/models/bus_stops_route_spec.rb
mboneil10: Alright

spec/models/bus_stops_route_spec.rb
mboneil10: Agreed. I wanted the `main_variant` to be the only variant, and thus contain the most stops already in order.

dfaulken: Okay. That's not quite what this accomplishes, though.

My thinking was that we wanted a scenario where there was another variant whose stops were a proper subset of the main variant, to make sure that the main variant wouldn't get reordered as a result.

app/models/bus_stops_route.rb
mboneil10: I think they're fine since they are descriptive.

app/models/bus_stops_route.rb
mboneil10: I think I'll go with `if stop == other_variant.first`.

spec/models/bus_stops_route_spec.rb
mboneil10: I don't really know how to be more descriptive other than `sequences stops correctly`

app/models/bus_stops_route.rb
mboneil10: Ok, I did that.

spec/models/bus_stops_route_spec.rb
dfaulken: Well, I mean that's the method's job. It sequences stops. So you could write that for any test description. It's more about what it does specifically given the circumstances.

So in this case, no stop is in the longest variant, what does it do? It adds the entire variant to the end of the longest variant.

Etc.

mboneil10: Ah, got it. Sorry wasn't on the same page. Thanks for clarifying.

lib/tasks/route_import.rake
```diff
@@ -6,52 +6,29 @@ require 'csv'
...
-              sequences << sequence
-              BusStopsRoute.create!(
-                bus_stop: stop, route: route, direction: direction, sequence: sequence 
-              )
-            end
-          end
+    BusStopsRoute.import(stop_hash)
+    
+    Route.delete_all
+    BusStopsRoute.delete_all
```
jforgue: A database cleaner gem might be a good option here.

jforgue: Whoops nvm, thought I was looking at a spec file when I wrote this!

spec/models/bus_stops_route_spec.rb
```diff
@@ -7,11 +7,11 @@
     context 'same route and direction' do
       let(:route) { create :route }
       let(:stop_1) { create :bus_stop }
-      let(:valid_bsr) {
+      let(:valid_bsr) do
```
jforgue: A nit pick, but you don't need parenthesis here.

spec/models/bus_stops_route_spec.rb
```diff
@@ -35,4 +35,57 @@
...
+    let(:other_variant) { %w[A B C D E] }
+    context 'all stops are in longest variant' do
+      it 'preserves sequence of longest variant' do
+        resequence!
+        expect(@stop_hash[:route_dir]).to eql %w[A B C D E]
+      end
+    end
+    context 'stop (but not first stop) is not in longest variant' do
+      context 'one stop is in other variant' do
+        let(:other_variant) { %w[A F E] }
```
jforgue: Isn't this two stops in the other variant?

mboneil10: @jforgue is confused by my test descriptions

spec/models/bus_stops_route_spec.rb
```diff
@@ -35,4 +35,57 @@
...
+      end
+      context 'multiple stops are in other variant' do
+        let(:other_variant) { %w[F G H D E] }
+        it 'inserts those stops into longest variant before common stop' do
+          resequence!
+          expect(@stop_hash[:route_dir]).to eql %w[A B C F G H D E]
+        end
+      end
+    end
+    context 'no stop is in the longest variant' do
```
mboneil10: @jforgue recommends to say "no stop in other variant is in the longest variant"

mboneil10: So that there isn't as much confusion between variants

mboneil10:
it's tested. What's the worst that can happen @jforgue :P

PR #105, opened 2018-07-09
Add current user to exception

jforgue:


jforgue:
According to exception_notification's read me this should will show up in the email as a new section titled "Data", and should look something like this.
`------------------------------- Data:

* data: {:current_user=>
  #<User:0x007ff03c0e5860
   id: 3,
   email: "jane.doe@example.com", # etc...`

jforgue:
I'm not sure how to test an exception, but I did check that `request.env["exception_notifier.exception_data"]` was assigned in the view.

app/controllers/application_controller.rb
```diff
@@ -4,11 +4,17 @@ class ApplicationController < ActionController::Base
...
+  before_action :prepare_exception_notifier
 
   helper_method :format_datetime
   
   private
 
+  def prepare_exception_notifier
+    request.env["exception_notifier.exception_data"] = {
+        :current_user => current_user
+    }
```
dfaulken: Two concerns here:

1. Does this hash exist before? Do we want to merge rather than overwrite?
2. Hash rockets in flight

jforgue: If I don't initialize the hash then it returns nil in the view. So I don't think it exists before I assign it.

app/controllers/application_controller.rb
```diff
@@ -4,11 +4,17 @@ class ApplicationController < ActionController::Base
...
 
   helper_method :format_datetime
   
   private
 
+  def prepare_exception_notifier
+    request.env['exception_notifier.exception_data'] = {
+      current_user: current_user
+    }
+  end
```
mboneil10: We typically at a newline at the end of methods.

app/controllers/application_controller.rb
```diff
@@ -4,11 +4,17 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_action :authenticate_user!
   before_action :set_paper_trail_whodunnit
+  before_action :prepare_exception_notifier
```
mboneil10: Should this be a `before_action` or `before_filter`? I'm looking at the documentation and you've used `before_action`. Why this change?

jforgue: `before_filter` is no longer supported in rails.

app/controllers/application_controller.rb
jforgue: OK

app/controllers/application_controller.rb
dfaulken: That may not necessarily be the case in production, however. Still best to merge rather than overwrite, I think.

jforgue: I mean the key `exception_notifier.exception_data` doesn't exist as far as I know, so I get a `NoMethodError (undefined method `merge' for nil:NilClass)`.

dfaulken: We can write code that handles it existing, and not overwriting if it does, and if it doesn't, initializing it with our value.

app/controllers/application_controller.rb
```diff
@@ -7,13 +7,18 @@ class ApplicationController < ActionController::Base
...
   helper_method :format_datetime
-  
+
   private
 
   def prepare_exception_notifier
-    request.env['exception_notifier.exception_data'] = {
-      current_user: current_user
-    }
+    if request.env['exception_notifier.exception_data'].nil?
```
dfaulken: I would recommend something like

```ruby
hash ||= {}
hash.merge(stuff)
```

PR #106, opened 2018-07-09
Index specs

mboneil10:
In an attempt to test most of the site before rewrite/refactor

### DIFF
--- a/spec/controllers/bus_stops_controller_spec.rb
+++ b/spec/controllers/bus_stops_controller_spec.rb
@@ -3,14 +3,24 @@ require 'spec_helper'
 describe BusStopsController, type: :controller do
   let(:incorrect_route_number) { '-1' }
   describe 'by_status' do
+    let :submit do
+      get :by_status, params: { number: incorrect_route_number }
+    end
     context 'incorrect route in params' do
       it 'redirects to index page' do
+        submit
+        expect(response).to redirect_to :index
       end
       it 'displays a helpful message' do
+        submit
+        expect(flash[:notice]).to eq "Route #{incorrect_route_number} not found"
       end
     end
   end
   describe 'by_sequence' do
+    let :submit do
+      get :by_sequence, params: { number: incorrect_route_number }
+    end
     context 'incorrect route in params' do
       it 'redirects to index page' do
       end
--- a/spec/controllers/bus_stops_controller_spec.rb
+++ b/spec/controllers/bus_stops_controller_spec.rb
@@ -1,6 +1,7 @@
 require 'spec_helper'
 
 describe BusStopsController, type: :controller do
+  let(:incorrect_route_number) { '-1' }
   context 'incorrect route in params' do
     context 'viewing stops by status' do
       before :each do
--- a/spec/controllers/bus_stops_controller_spec.rb
+++ b/spec/controllers/bus_stops_controller_spec.rb
@@ -27,8 +27,12 @@ describe BusStopsController, type: :controller do
     end
     context 'incorrect route in params' do
       it 'redirects to index page' do
+        submit
+        expect(response).to redirect_to bus_stops_path
       end
       it 'displays a helpful message' do
+        submit
+        expect(flash[:notice]).to eq "Route #{incorrect_route_number} not found"
       end
     end
   end
--- a/spec/controllers/bus_stops_controller_spec.rb
+++ b/spec/controllers/bus_stops_controller_spec.rb
@@ -13,7 +13,7 @@ describe BusStopsController, type: :controller do
     context 'incorrect route in params' do
       it 'redirects to index page' do
         submit
-        expect(response).to redirect_to :index
+        expect(response).to redirect_to bus_stops_path
       end
       it 'displays a helpful message' do
         submit
--- a/spec/controllers/bus_stops_controller_spec.rb
+++ b/spec/controllers/bus_stops_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe BusStopsController, type: :controller do
+describe BusStopsController do
   let(:incorrect_route_number) { '-1' }
   let!(:user) { create :user }
   before :each do
--- a/spec/controllers/bus_stops_controller_spec.rb
+++ b/spec/controllers/bus_stops_controller_spec.rb
@@ -2,6 +2,10 @@ require 'spec_helper'
 
 describe BusStopsController, type: :controller do
   let(:incorrect_route_number) { '-1' }
+  let!(:user) { create :user }
+  before :each do
+    when_current_user_is user
+  end
   describe 'by_status' do
     let :submit do
       get :by_status, params: { number: incorrect_route_number }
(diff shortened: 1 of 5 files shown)


mboneil10:
Since we're here, I don't understand when [this line](https://github.com/umts/stop-project/blob/master/app/controllers/bus_stops_controller.rb#L18) or [this line](https://github.com/umts/stop-project/blob/master/app/controllers/bus_stops_controller.rb#L31) will run since we use those methods after selecting a`route` from the dropdown.

jforgue:
Shouldn't we test the actual order of the two index pages? Or were you planning on doing that in a different PR?

dfaulken:
> Since we're here, I don't understand when [this line](https://github.com/umts/stop-project/blob/master/app/controllers/bus_stops_controller.rb#L18) or [this line](https://github.com/umts/stop-project/blob/master/app/controllers/bus_stops_controller.rb#L31) will run since we use those methods after selecting a`route` from the dropdown. And the dropdown consists of `options_for_select(Route.pluck(:number))`?
>
> Could it help in a scenario when a route is deleted but still shows up in the dropdown?

It wasn't always a dropdown, is the short answer. However, let's keep in mind that there's not necessarily a hard link between how your preceding view is constructed, and what params the controller will receive. "Don't trust params from the scary internet", as dhh would say.

mboneil10:
@jforgue I already have a test for [viewing stops by status](https://github.com/umts/stop-project/blob/master/spec/system/bus_stops/viewing_bus_stops.rb). I was going to add another one, viewing stops by sequence. If you think I should put those two tests in the index spec I can. But I thought if I could separate some tests out, I should.

spec/system/index_spec.rb
```diff
@@ -33,10 +33,6 @@
                                     text: 'Stop -1 not found'
     end
   end
-  context 'without completing stop id' do
```
mboneil10: Don't know why I thought this was a functionality of the index page. Just to be sure, this isn't something that we were hoping to implement, correct?

mboneil10:
Hi friends, this is ready for re-review

jforgue:
Ah, I confused our index page with the pages that actually display the stops. My bad. Although that does make me question our layout a little bit.

mboneil10:
? @jforgue how so?

spec/system/index_spec.rb
```diff
@@ -87,19 +84,61 @@
...
+      before :each do
+        click_link 'View by route order'
+      end
+      it 'redirects to bus stops by sequence' do
+        expect(page).to have_content route.number
+        expect(page.current_path).to end_with by_sequence_bus_stops_path
+      end
+    end
+  end
+  context 'incorrect route in params' do
```
jforgue: I would rename this as system tests are all about user interaction and users can't directly change params.

spec/system/index_spec.rb
```diff
@@ -30,18 +31,15 @@
...
-    it 'autofills' do
+                                    text: "Stop #{incorrect_stop_id} not found"
     end
   end
 end
 
 describe 'searching for a bus stop by stop name' do
   let(:user) { create :user }
   let!(:bus_stop) { create :bus_stop }
+  let(:incorrect_stop_name) { 'stahp' }
```
jforgue: Approved for a high quality joke.

jforgue:
I would think an index displays data and the search would be a separate page.

mboneil10:
Ok, so what do you mean? What tests in particular would you suggest I move? Or are you saying something else about index versus search?

spec/system/index_spec.rb
mboneil10: Thanks. Idk if the word `incorrect` relays exactly what I want it to.

spec/system/index_spec.rb
dfaulken: Good question, actually, as we've established that this isn't something that can really happen as a result of user interaction.

I would argue that a controller test might be more appropriate here, and in general when testing any inputs that a controller can receive but which the corresponding view cannot produce.

mboneil10: Alright, will do.

spec/system/index_spec.rb
```diff
@@ -67,18 +65,17 @@
...
-                                    text: 'Stop 1234 not found'
+                                    text: "Stop #{incorrect_stop_name} not found"
     end
   end
   context 'without completing stop name' do
-    let!(:bus_stop) { create :bus_stop, name: 'Mill Valley Apartments' }
     it 'autofills' do
-      fill_in 'Enter stop name', with: 'Mill Valley'
-      find('div', text: 'Mill Valley Apartments').click
+      fill_in 'Enter stop name', with: bus_stop.name.chars.first
```
dfaulken: I'm confused, why the first character?

mboneil10: Because I want to test that autofill works. In order to do that, I fill in the `text_field` with the first letter of the stop name: `M`. Then `autocomplete` will create a `div` with `Mill Valley Apartments`.

dfaulken: Okay, I see. Perhaps a comment to that effect might be useful - or, making the reasonable decision that most people who read this will be smarter than Dave.

mboneil10: I'll write in a comment.

spec/system/index_spec.rb
```diff
@@ -67,18 +65,18 @@
...
-                                    text: 'Stop 1234 not found'
+                                    text: "Stop #{incorrect_stop_name} not found"
     end
   end
   context 'without completing stop name' do
-    let!(:bus_stop) { create :bus_stop, name: 'Mill Valley Apartments' }
     it 'autofills' do
-      fill_in 'Enter stop name', with: 'Mill Valley'
-      find('div', text: 'Mill Valley Apartments').click
+      # use the first character of the stop name so it's incomplete
```
mboneil10: What do people think of this comment?

dfaulken: ... so that it can be completed by the autocomplete, right?

mboneil10: Dammit thanks dave

spec/controllers/bus_stops_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+
+describe BusStopsController, type: :controller do
```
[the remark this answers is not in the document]
dfaulken: It is not:

https://github.com/umts/stop-project/blob/28f3dd5f437b6998a63fd314137353d1c9418a72/spec/spec_helper.rb#L36

spec/controllers/bus_stops_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      get :by_status, params: { number: incorrect_route_number }
+    end
+    context 'incorrect route in params' do
+      it 'redirects to index page' do
+        submit
+        expect(response).to redirect_to bus_stops_path
+      end
+      it 'displays a helpful message' do
+        submit
+        expect(flash[:notice]).to eq "Route #{incorrect_route_number} not found"
```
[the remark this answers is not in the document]
dfaulken: One could follow an old bookmark generated before the change, for example.

spec/controllers/bus_stops_controller_spec.rb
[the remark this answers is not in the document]
mboneil10: whoops I generated that file.

Issue #101, opened 2018-06-27
Use GTFS export for routes


sherson:
@dfaulken is investigating some issues with the current export (#97 #98). I mentioned to him that another approach possibly worth considering is using the GTFS export:

- PVTA's current file: http://www.pvta.com/g_trans/ 
- Archive: https://transitfeeds.com/p/massdot/104
- Specs: https://developers.google.com/transit/gtfs/reference/

We would need to confirm with SATCo and VATCo which dates the feed files need to cover in order to include all the routes.

The upside being that we would not have to create/troubleshoot/maintain our own control file.

dfaulken:
Should we get a data sample now so we can start development, in anticipation of this being the path forward? Or should we wait until we have approval from VATCo/SATCo/PVTA?

sherson:
We don't need approval from VATCo/SATCo/PVTA for this; it's up to us (I'm positive they won't object to using an already produced/supported/hosted source). 

To me it's appealing since (along with not having to create/troubleshoot/maintain our own control file) we're already familiar with GTFS for other things anyway, but I don't feel strongly about it. Do you have any opinion? 

Based on the contents of `calendar.txt` and `calendar_dates.txt`, the [2018-05-01](https://transitfeeds.com/p/massdot/104/20180501) file looks good (assuming the date range of 5/6-12 is in fact good, which is what we were told).

sherson:
So to answer your question a bit more directly, I think we should develop this, unless anyone has any objections.

dfaulken:
@mboneil should we roll into the open PR's rake task, or address the algorithm separately before taking on a new data source?

mboneil10:
Oh sorry @dfaulken I didn't see this comment four days ago.

mboneil10:
I think since we should address the algorithm separately before taking on a new data source. That's just my impression, in order to keep changes as clear as possible. What do you think @dfaulken?

Anbranin:
@sherson wants me to work on this unless someone else wants to tackle it. I'm going to wait until you all address whatever problems you were having with the algorithm, though. 
Or perhaps I can at least start it. Let me know. I'd prefer someone else assign me to the issue, however. I'm available when you need me after I finish up a couple things today.

PR #111, opened 2018-07-21
Gtfs import

Anbranin:
Everything seems to be working. The tests pass, I did some manual checking of routes.
The only issue is that "direction" is now "0" or "1". I did not have time to add headsigns so that we could use those instead of displaying the direction, but it's better that that go in a different PR.

### DIFF
--- a/app/models/bus_stops_route.rb
+++ b/app/models/bus_stops_route.rb
@@ -18,13 +18,13 @@ class BusStopsRoute < ApplicationRecord
     # The route direction array key isn't used because it doesn't matter here.
     # Since stop_hash is used for creating routes and bus_stops_routes, though,
     # we need to keep those keys.
-    stop_hash.each_pair do |route_dir, direction_variants|
-      variants = direction_variants.values
-      longest_variant = variants.max_by(&:length)
-      other_variants = variants - longest_variant
-      other_variants.each do |other_variant|
-        other_variant.each.with_index do |stop, sequence|
-          unless longest_variant.include? stop
+    stop_hash.each_pair do |route_dir, direction_trips|
+      trips = direction_trips.values.uniq.map(&:compact).map(&:uniq)
+      longest_trip = trips.max_by(&:length)
+      other_trips = trips - [longest_trip]
+      other_trips.each do |trip|
+        trip.each.with_index do |stop, sequence|
+          unless longest_trip.include? stop
             # We can't rely on the previous stop here
             # like we do in the corresponding else statement.
             if stop == other_variant.first
--- a/app/models/bus_stops_route.rb
+++ b/app/models/bus_stops_route.rb
@@ -12,9 +12,9 @@ class BusStopsRoute < ApplicationRecord
   
   # This method is used for importing a csv of routes. The input stop_hash
   # has a route and direction array pointing to trips with stops (ordered
-  # by sequence). import combines all stops per route direction, and
-  # sequences those stops accordingly.
-  def self.import(stop_hash)
+  # by sequence). establish_sequences combines all stops per route direction, 
+  # and sequences those stops accordingly.
+  def self.establish_sequences(stop_hash)
     # The route direction array key isn't used because it doesn't matter here.
     # Since stop_hash is used for creating routes and bus_stops_routes, though,
     # we need to keep those keys.
--- a/app/models/bus_stops_route.rb
+++ b/app/models/bus_stops_route.rb
@@ -10,15 +10,15 @@ class BusStopsRoute < ApplicationRecord
   validates_uniqueness_of :sequence, scope: %i[route direction]
   validates_uniqueness_of :bus_stop, scope: %i[route direction]
   
-  # This method is used for importing a csv of routes. The input stop_hash
+  # This method is used for importing a csv of routes. The input stop_data
   # has a route and direction array pointing to trips with stops (ordered
   # by sequence). establish_sequences combines all stops per route direction, 
   # and sequences those stops accordingly.
-  def self.establish_sequences(stop_hash)
+  def self.establish_sequences(stop_data)
     # The route direction array key isn't used because it doesn't matter here.
-    # Since stop_hash is used for creating routes and bus_stops_routes, though,
+    # Since stop_data is used for creating routes and bus_stops_routes, though,
     # we need to keep those keys.
-    stop_hash.each_pair do |route_dir, direction_trips|
+    stop_data.each_pair do |route_dir, direction_trips|
       trips = direction_trips.values.uniq.map(&:compact).map(&:uniq)
       longest_trip = trips.max_by(&:length)
       other_trips = trips - [longest_trip]
@@ -54,7 +54,7 @@ class BusStopsRoute < ApplicationRecord
       # condenses all trip and stop data into one array of stop ids,
       # representing the ordered path of all possible stops on that
       # route in one particular direction.
-      stop_hash[route_dir] = longest_trip
+      stop_data[route_dir] = longest_trip
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/bus_stops_route_creation.rake
```diff
@@ -0,0 +1,57 @@
...
+            route: route, direction: direction, bus_stop: stop, sequence: sequence
+          )
+        end
+      end
+    end
+    ActionMailer::Base.mail(
+      to: 'transit-it@admin.umass.edu',
+      subject: 'missing trips in GTFS export',
+      body: "The following trips were missing from trips.txt: #{missing_trips.join(',')}"
+    )
```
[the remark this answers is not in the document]
Anbranin: > Why email?
I've been working a lot on automated tasks recently so did this unthinkingly. That's all. Will `puts` instead!

Anbranin:
I should note that I realize filtering by trips is inefficient, but these files do not contain variants. What they do contain is a shape_id, which would function perfectly in place of a variant _if_ all the trips actually had a shape_id, which they don't. 
Using the trip_id was the only viable choice.

dfaulken:
>  I realize filtering by trips is inefficient, but these files do not contain variants. 

True, but my point was that we perform the algorithm on every single trip, regardless of whether we have already processed a trip with an identical sequence of stops. This seems inefficient.

Anbranin:
Oh, that seems like an easy thing to change! But which algorithm do you mean?

dfaulken:
> But which algorithm do you mean?

Sorry — the model method.

Anbranin:
I thought that's what I was doing with line 22 😿 so I'm probably not seeing what you're talking about.

Anbranin:
You're correct, but I'm using both:
`trips = direction_trips.values.uniq.map(&:compact).map(&:uniq)`
Does this resolve your question?
Should I make these into two lines and add some comments so it's clearer what's going on? If the code is that unreadable I should work on it, whatever you think.

dfaulken:
Fine for now. Sorry for the confusion. That's fine for functionality. I'll do a style review this PM.

app/models/bus_stops_route.rb
```diff
@@ -18,43 +18,43 @@ def self.import(stop_hash)
...
+              previous_stop = trip[sequence - 1]
+              previous_stop_index_in_longest_trip = longest_trip.index(previous_stop)
               # Array#insert inserts before the given index.
-              longest_variant.insert(previous_stop_index_in_longest_variant + 1, stop)
+              longest_trip.insert(previous_stop_index_in_longest_trip + 1, stop)
             end
           end
         end
       end
       # overwrites the data we don't need, that is, the variants and corresponding
```
dfaulken: Still reads variant (and line 55)

lib/tasks/bus_stops_route_creation.rake
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+namespace :bus_stops_route do
+  task :import => :environment do
```
dfaulken: No hashrockets please

lib/tasks/bus_stops_route_creation.rake
```diff
@@ -0,0 +1,53 @@
...
+    stop_hash = {}
```
dfaulken: No names of variable types in the variable name please

lib/tasks/bus_stops_route_creation.rake
```diff
@@ -0,0 +1,53 @@
...
+    missing_trips = []
+
+    CSV.foreach('trips.txt', headers: true) do |row|
+      route_number = row['route_id']
+      direction = row['direction_id']
+      trip_id = row['trip_id']
+      unless route_number.blank?
```
dfaulken: Perhaps a comment explaining why this might be (I know, but future others may not)

lib/tasks/route_import.rake
```diff
@@ -6,31 +6,10 @@ require 'csv'
...
-      Route.find_or_create_by! number: route.strip
-      stop_ids.each.with_index(1) do |stop_id, sequence|
-        stop = BusStop.find_by(hastus_id: stop_id)
-        if stop.present?
-          BusStopsRoute.create!(route: route, direction: direction, bus_stop: stop, sequence: sequence)
-        end
-      end
+    CSV.foreach(args[:csv_file], headers: true, col_sep: ';') do |row|
+      route_number = row['route_id']
+      Route.find_or_create_by! number: route_number.strip
```
dfaulken: Since we're not using my latest OIG to import BSRs, we can use it to just get a flat listing of routes, and then we can get description back. I don't think it's out of scope to roll that into a PR where we use GTFS to import BSRs, but open to being overruled on that.

lib/tasks/bus_stops_route_creation.rake
Anbranin: Wasn't me

lib/tasks/bus_stops_route_creation.rake
Anbranin: I don't know if I can explain it. There are visually no blank rows in this CSV, but I'm on a blank row when `route_number` is blank (everything else is just `nil`, but `route_number1` is a blank string.

Anbranin: It only happens once btw.

Anbranin: What if it's a hidden character type deal, like when the zero-width space character was causing me all sorts of headaches?

lib/tasks/bus_stops_route_creation.rake
```diff
@@ -0,0 +1,53 @@
...
+        stop_hash[[route_number, direction]][trip_id] ||= []
+      end
+    end
+
+    CSV.foreach('stop_times.txt', headers: true) do |row|
+      trip_id = row['trip_id']
+      sequence = row['stop_sequence'].to_i
+      stop_id = row['stop_id']
+
+      route = stop_hash.select do |route_direction, trips|
```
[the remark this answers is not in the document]
Anbranin: So much cleaner, thanks.

lib/tasks/bus_stops_route_creation.rake
```diff
@@ -0,0 +1,53 @@
...
+        trips.keys.include? trip_id
+      end
+      if route.empty?
+        missing_trips << trip_id
+      else
+        route.values.first[trip_id][sequence] = stop_id
+      end
+    end
+
+    BusStopsRoute.import(stop_hash)
```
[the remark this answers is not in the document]
Anbranin: I'm cool with `establish_sequences` because that's what's happening. I'll change it to that for now. @mboneil10?

lib/tasks/route_import.rake
Anbranin: You've used one too many acronyms, my friend. I'd be happy to do whatever it is. I have two questions:
1. What is OIG?
2. By description, do you mean "direction"? If so, I see that yeah, there's direction there, but what kind of gymnastics do I have to do to figure out which direction from this file goes with which longest_trip on that other one? I can't fathom how to match them easily. If you can please share. Or go for it.

dfaulken: This is more an item for @mboneil10. Can table for this PR and do another subsequent PR.

lib/tasks/bus_stops_route_creation.rake
dfaulken: Totally understood

lib/tasks/route_import.rake
Anbranin: Sounds good.

Anbranin:
@dfaulken how does this look now? I don't want this to get buried in the sands of time.

lib/tasks/bus_stops_route_creation.rake
Anbranin: Ah ok

Anbranin:
SANDS OF TIME

mboneil10:
Closes #101

PR #117, opened 2018-08-02
Save and next bus stop

### LINKED ISSUE #67: Save Stop - Option to remember route
Feature request from Leigh:

The workflow for the users is typically to:

1. Select a route
2. Pick a stop from that route list
3. Mark as completed and hit Save

Currently, that sends the user back to the select stop screen, from which they need to reselect the route. It would be convenient if the user could go back to the route they'd previously selected.

This could be accomplished in various ways. Leigh's first choice is to have a button that goes to the next incomplete stop on the route previously selected. Another option is to have a button that simply goes back to the previous page, or a button that saves without changing pages (and the user can simply tap the browser's back button).

Not high priority.

jforgue:
Closes #67 

screen shot 2018-08-02 at 11 32 05 am


app/controllers/bus_stops_controller.rb
```diff
@@ -81,7 +85,18 @@ def outdated
   def update
     @stop.assign_attributes stop_params
     @stop.decide_if_completed_by current_user
-    if @stop.save
+    if params[:commit] == 'Save and next'
+      route_id = session[:route_id]
+      next_stop = Route.find_by_id(route_id)
```
mboneil10: I thought `find_by_x` was considered a no-go?

app/controllers/bus_stops_controller.rb
```diff
@@ -81,7 +85,18 @@ def outdated
...
+                       .next_stop_in_sequence(@stop, route_id, session[:direction])
+      flash[:notice] = 'Bus stop was updated.'
+      if next_stop
+        redirect_to(edit_bus_stop_url(next_stop.hastus_id))
+      else
+        route_number = Route.find(route_id).number
+        redirect_to(by_sequence_bus_stops_url(number: route_number))
```
mboneil10: I think this is a good choice

app/models/route.rb
```diff
@@ -3,4 +3,17 @@ class Route < ApplicationRecord
   has_many :bus_stops_routes
   has_many :bus_stops, through: :bus_stops_routes
   default_scope { order :number }
+
+  def stops_by_sequence(route_id, direction)
+    BusStop
```
mboneil10: why is `BusStop` all alone on this line? It's not so readable in my opinion. I'd suggest to at least have `BusStop.joins(:bus_stops_routes)` for the first line of this query.

app/models/route.rb
```diff
@@ -3,4 +3,17 @@ class Route < ApplicationRecord
...
+      .joins(:bus_stops_routes)
+      .where(bus_stops_routes: { route_id: route_id, direction: direction })
+      .order('sequence')
+  end
+
+  def next_stop_in_sequence(stop, route_id, direction)
+    stops = stops_by_sequence(route_id, direction)
+    index = stops.find_index(stop) + 1
```
mboneil10: I'd suggest to rename this variable. Yes, it's clear that it is an `index`. But why does that matter? Perhaps it could be named `index_of_next_stop`? Or something along those lines.

app/views/bus_stops/_form.haml
```diff
@@ -48,4 +48,7 @@
         %td= f.check_box :completed, class: 'triple-zoom'
     %tr
       %td= submit_tag 'Save stop', class: 'btn btn-primary btn-lg btn-block submit'
+      - unless session[:route_id].nil?
```
mboneil10: Alternatively, `-if session[:route_id]`, right? Curious why this was chosen instead.

app/controllers/bus_stops_controller.rb
```diff
@@ -21,6 +22,7 @@ def by_sequence
   end
 
   def by_status
+    session.delete(:route_id)
```
dfaulken: Why bother, out of curiosity (here as well as below)? If it only matters that it's there when we need it, does it matter if it's there when we don't?

app/controllers/bus_stops_controller.rb
```diff
@@ -81,7 +85,18 @@ def outdated
   def update
     @stop.assign_attributes stop_params
     @stop.decide_if_completed_by current_user
-    if @stop.save
+    if params[:commit] == 'Save and next'
+      route_id = session[:route_id]
```
dfaulken: Session is like params - we should be prepared for values not to be there.

app/views/bus_stops/_form.haml
```diff
@@ -48,4 +48,7 @@
         %td= f.check_box :completed
     %tr
       %td= submit_tag 'Save stop', class: 'btn btn-primary btn-lg btn-block submit'
+      - if session[:route_id]
```
dfaulken: Generally should use instance variables to control view logic whenever possible.

app/controllers/bus_stops_controller.rb
dfaulken: * Particularly when we infer from the parameters that a particular session value will be present.

app/controllers/bus_stops_controller.rb
dfaulken: See my general comment - I would be interested in determining whether sessions are actually necessary for this feature, and then returning to the best way to engineer it.

jforgue:
@dfaulken passing a BSR was my original plan, but BSRs don't have ids so I would have to pass at least two params to identify the BSR anyways. Also, each stop has it's own BSR when sequenced. Of course, I didn't think to write a model method.

app/controllers/bus_stops_controller.rb
```diff
@@ -81,7 +81,21 @@ def outdated
   def update
     @stop.assign_attributes stop_params
     @stop.decide_if_completed_by current_user
-    if @stop.save
+    if params[:commit] == 'Save and next' && params[:route_id]
+      route_id = params[:route_id]
+      next_stop = Route.find(route_id)
+                       .next_stop_in_sequence(@stop, route_id, params[:direction])
+      flash[:notice] = 'Bus stop was updated.'
```
Anbranin: No it wasn't.

app/models/route.rb
```diff
@@ -3,4 +3,16 @@ class Route < ApplicationRecord
   has_many :bus_stops_routes
   has_many :bus_stops, through: :bus_stops_routes
   default_scope { order :number }
+
+  def stops_by_sequence(route_id, direction)
```
dfaulken: These are both instance methods. Why do we pass the route ID around both methods?

app/controllers/bus_stops_controller.rb
```diff
@@ -83,13 +83,35 @@ def update
...
     else
       flash[:errors] = @stop.errors.full_messages
       render 'edit'
     end
   end
 
+  def edit
+    if params[:route_id]
+      @route = Route.find(params[:route_id])
+      @direction = params[:direction]
```
dfaulken: Let's `require` this parameter.

app/controllers/bus_stops_controller.rb
```diff
@@ -83,13 +83,35 @@ def update
     @stop.decide_if_completed_by current_user
     if @stop.save
       flash[:notice] = 'Bus stop was updated.'
-      redirect_to bus_stops_path
+      if params[:commit] == 'Save and next' && params[:route_id]
+        route_id = params[:route_id]
+        next_stop = Route.find(route_id)
+                         .next_stop_in_sequence(@stop, route_id, params[:direction])
+        if next_stop
+          redirect_to(action: 'edit',
```
dfaulken: Shouldn't need the parentheses around either `redirect_to`, which I mention because we don't usually write them.

app/models/route.rb
jforgue: If we didn't we would have to call both methods in the controller. Alternatively I could make this one method.

dfaulken: My point is that these are instance methods, and so you already know what the route is. It's `self`. You shouldn't need to pass it as an argument.

jforgue: Oh, I see.

app/controllers/bus_stops_controller.rb
```diff
@@ -81,14 +81,33 @@ def update
...
     if @stop.save
       flash[:notice] = 'Bus stop was updated.'
-      redirect_to bus_stops_path
+      if params[:commit] == 'Save and next' && params[:route_id]
+        route_id = params[:route_id]
+        next_stop = Route.find(route_id)
+                         .next_stop_in_sequence(@stop, params[:direction])
+        if next_stop
+          redirect_to action: 'edit',
+                      id: next_stop.hastus_id,
```
mboneil10: Should this be `hastus_id` or a `bus_stop` `id`?

app/models/route.rb
```diff
@@ -3,4 +3,16 @@ class Route < ApplicationRecord
   has_many :bus_stops_routes
   has_many :bus_stops, through: :bus_stops_routes
   default_scope { order :number }
+
+  def stops_by_sequence(direction)
```
mboneil10: I appreciate how you used a method here instead of shoving that query into the method below.

app/controllers/bus_stops_controller.rb
jforgue: The path relies on `hastus_id` for some reason.

dfaulken: When you can show a meaningful identifier, show a meaningful identifier, was more or less my thinking.

app/models/route.rb
```diff
@@ -3,4 +3,15 @@ class Route < ApplicationRecord
...
+    bus_stops.where(bus_stops_routes: { direction: direction })
+             .order('bus_stops_routes.sequence')
```
dfaulken: OOC, does this not work as just `order(:sequence)`? I think BSR is the only table with a `sequence` column.

app/controllers/bus_stops_controller.rb
dfaulken: This should've occurred to me before - can we use the path helper here, like in `by_sequence.haml:13`?

app/models/route.rb
werebus: Huh, who knew?  From the Rails 5 documentation:

```ruby
User.order(:name) # SELECT "users".* FROM "users" ORDER BY "users"."name" ASC
User.order('name') # SELECT "users".* FROM "users" ORDER BY name
```

PR #118, opened 2018-08-02
GTFS import routes

### LINKED ISSUE #101: Use GTFS export for routes

dfaulken:
Closes #101.

### DIFF
--- a/lib/tasks/route_import.rake
+++ b/lib/tasks/route_import.rake
@@ -7,9 +7,11 @@ require 'csv'
 namespace :routes do
   task :import, [:csv_file] => :environment do |_, args|
     Route.delete_all
-    CSV.foreach(args[:csv_file], headers: true, col_sep: ';') do |row|
-      route_number = row['route_id']
-      Route.find_or_create_by! number: route_number.strip
+    CSV.foreach(args[:csv_file], headers: true) do |row|
+      number = row['route_short_name']
+      description = row['route_long_name']
+      route = Route.find_or_create_by! number: number
+      route.update! description: description
     end
   end
 end


mboneil10:
(aka I just ran it)

werebus:
One gotcha to mention with using the GTFS routes file is that it only contains the routes that run in the time span it's exported for. So, the current one is missing routes that don't run in the summer (e.g. 38, 39, 34, 35)

dfaulken:
@sherson didn't we assess that for this approach? Or was that when we were thinking about extracting directly from HASTUS?

mboneil10:
huh

sherson:
Yep https://github.com/umts/stop-project/issues/101#issuecomment-401469662

sherson:
There were some communications elsewhere about this, but to clarify here for posterity: PVTA advised us to use the week of 2018-05-06 to 2018-06-12, which is in the archive at https://transitfeeds.com/p/massdot/104/20180501. When we do this process in the distant future, we'll need to confirm with PVTA what week we should import, and then determine if we can use the current GTFS file, or an archived one based on the calendar files. See links in first comment at #101.

PR #124, opened 2018-08-07
route number is not the same as route_id.

### LINKED ISSUE #121: Routes missing stops
@mboneil10 provided a list of routes without stops according to:

`Route.includes(:bus_stops).where(bus_stops: {id: nil}).pluck(:number).uniq`

> "10s", "NE", "OWL", "P20E", "P21E", "PV", "T1", "T2", "T20", "T21", "T23", "T24", "T3", "T4", "T5", "T92", "WS"

Confirmed that P21E does not have any stops associated with it in the app (through the UI). Looked at archived GTFS file (that was used for the last import):

https://transitfeeds.com/p/massdot/104/20180501

routes.txt has:

```
921,P21E
```

trips.txt has:

```
921,S1803-AMD-Weekday-53,2609250-S1803-AMD-Weekday-53 
```

stop_times.txt has:

```
2609250-S1803-AMD-Weekday-53,06:27:31,06:27:31,387
```

This indicates that stop 387 should be associated with P21E. Any thoughts as to why it’s not?

Related notes:

* Stop 387 is on P21E in Avail and at [maps.google.com](https://goo.gl/maps/sZMF66ebbgw) (based on the current GTFS file). 
* I’m not sure if this actually matters, but calendar.txt does have that service_id defined:

```
S1803-AMD-Weekday-53,1,1,1,1,1,0,0,20180622,20180622
```

Anbranin:
We'd been using the route_ID as the route_number, assuming they were the same thing.
The route_number and route_ID are both available in routes.txt, but that is the only place that ties them together. The route_ID is referenced in all the other files, but never saved to the route. Therefore, I thought the solution was to just combine all the tasks in one file so I could easily save the data temporarily about which routes had which route_ID. It doesn't noticeably affect the speed of the task.

closes #121


dfaulken:
I'm a little confused on the problem statement, especially given that there are still 'troubleshooting' comments in this code.

Could we break down a little more clearly what we were doing before, which data rows it didn't work with, and what we're doing now?

Anbranin:
I forgot to remove the comments - they were copied and pasted from the issue (it was @sherson's example) so I had an example of a route that had no stops and why it _should_have had stops, based on the data. I put a binding.pry in and stepped through the import of just that route's data, and it was immediately quite obvious what my mistake was earlier. I'll step through that example here:

1. In the route import, the `route_number` is `P21E` and the `route_id` is `921`.  We discard the `route_id`, as you can see [here](https://github.com/umts/stop-project/blob/master/lib/tasks/route_import.rake#L11-L13) there's no mention of it. At this point, the relationship between "921" and "P21E" is lost and we're done with that task.
2. In the task that imports the rest of the data, we come first to `trips.txt`. [This line](https://github.com/umts/stop-project/blob/master/lib/tasks/bus_stops_route_creation.rake#L11) indicates with a variable name that the author (me lol) thought that `route_id` was the same as `route_number`, but let's look at an example line from the same route, P21E:
`921,S1803-AMD-Weekday-53,2609250-S1803-AMD-Weekday-53`. In this line, `route_id` is `921`, and `trip_id` is `2609250-S1803-AMD-Weekday-53`. The hash variable that keeps track of data imported now contains 921, 2609250-S1803-AMD-Weekday-53, and a direction. (0 or 1). 
3. Now we come to the `stop_times.txt` which contains sequencing data and is the final CSV import. Here an example line for this route is 
`2609250-S1803-AMD-Weekday-53,06:27:31,06:27:31,387` which contains, along with some data we discard, `trip_id`, which we use to match with the data already in the hash to store this trip data with the correct route/direction combo, the `sequence` and the `stop_id`. After this file is imported, we send the hash into `establish_sequences` which does what it says on the tin and has no bearing on the troubleshooting here. The hash comes out of that method much cleaner-looking but still with the keys of `[921, 0]` and `[921, 1]` referencing the route `P21E`.
4. Now we save the BusStopRoutes. You can tell from the very [beginning of the last block](https://github.com/umts/stop-project/blob/master/lib/tasks/bus_stops_route_creation.rake#L38) that the author still thinks that `921` is going to be a route number, and helpfully, the [next line](https://github.com/umts/stop-project/blob/master/lib/tasks/bus_stops_route_creation.rake#L39) saves a route with that number even if it's not found (didn't get imported in the routes import). Since 921 didn't get imported, it's created, and of course `P21E` didn't appear in any of those other files. So poor P21E never gets any stops associated with it, but its clone 921 does.

The reason this wasn't noticed is because all or at least most of the routes the author was familiar with (the UMTS routes.. ahem..) had the same route_id as route_number. I had no idea there wasn't a "route 921" so that threw no red flags, didn't look for routes with no stops (didn't think to), and the tests unfortunately are only as good as the data they're given. You can only test the scenarios the test author thinks of, after all.

Is that clear?

dfaulken:
Wonderfully clear, thank you!

PR #126, opened 2018-08-09
Return name instead of user for completed_by

jforgue:


dfaulken:
This was a pretty minor change, but it would've been nice for someone on the project team to be able to review.

Anbranin:
As I stated in another issue, but bears repeating here:
@sherson asked me if I could deploy as soon as this was done because Krystal had asked him for the CSV.

PR #128, opened 2018-08-14
Add Hastus id to bus stop name

### LINKED ISSUE #122: Searching by stop name when not unique
PVTA stop names are not always unique (there are many duplicates and some triplicates). Currently, searching by stop name (rather than stop ID) always returns the first stop name on the list. For example, choosing the name for one of these...

name | id
--|--
Plumtree / Mayflower | 1590
Plumtree / Mayflower | 1596
Plumtree / Mayflower | 1607

...always returns 1590.

The best solution is for PVTA to use unique names... we'll ask them if that's planned.

jforgue:
Closes #122


jforgue:
screen shot 2018-08-14 at 4 06 06 pm

jforgue:
screen shot 2018-08-14 at 4 06 24 pm
 We may want to remove the `hastus_id` display from the `by_sequence` view.

jforgue:
@dfaulken Do you think the way I'm handling this is okay and it should be abstracted to the model, or would You recommend I try something else in the model?

dfaulken:
Instance method would be most straightforward.

app/controllers/bus_stops_controller.rb
```diff
@@ -53,7 +54,8 @@ def manage
   end
 
   def name_search
-    stop = BusStop.find_by name: params.require(:name)
+    name = params[:name][/[^(]+/].strip
```
jforgue: Should this also be a model method?

app/controllers/bus_stops_controller.rb
```diff
@@ -5,6 +5,7 @@ class BusStopsController < ApplicationController
   def autocomplete
     stops = BusStop.where 'lower(name) like ?',
                           "%#{params.require(:term)}%"
+    stops.each &:add_hastus_id_to_name
```
dfaulken: Definitely parentheses here.

app/controllers/bus_stops_controller.rb
dfaulken: Yes please.

app/models/bus_stop.rb
```diff
@@ -245,6 +245,10 @@ def decide_if_completed_by(user)
     end
   end
 
+  def add_hastus_id_to_name
+    self.name = name + " (#{hastus_id})"
```
dfaulken: I don't agree that we should be editing the objects. Again - this is a *presentational* concern, not a *data* concern. Let's have this be a presentational method only, not one that edits the objects.

mboneil10:
Had to remove my request since I will be working on this PR.

app/controllers/bus_stops_controller.rb
jforgue: This is how I would make this a model method `name = BusStop.remove_hastus_id_from_name(params[:name])`. Since `params[:name]` doesn't actually belong to a BusStop, is that the right thing to do?

app/models/bus_stop.rb
```diff
@@ -245,6 +245,10 @@ def decide_if_completed_by(user)
     end
   end
 
+  def add_hastus_id_to_name
+    name + " (#{hastus_id})"
```
nawilliams: Might as well string interpolate the whole thing.
`"#{name} (#{hastus_id})"`

Function should probably be renamed, add makes it sound like it makes a permanent change to name rather than just returning them combine. Something like `name_with_hastus_id` is more clear what it does.

dfaulken:
@nawilliams I can't explicitly request you as a review on this project, but if you happen to be available to look over it again, that would be great.

PR #134, opened 2018-08-29
Broken migrations

mboneil10:
Didn't need to make new migration for creating fields -- it's already [there](https://github.com/umts/stop-project/blob/rewrite/db/migrate/20180614232652_create_fields.rb)

### DIFF
--- a/db/migrate/20180608191700_add_choices_to_field.rb
+++ b/db/migrate/20180608191700_add_choices_to_field.rb
@@ -1,5 +1,0 @@
-class AddChoicesToField < ActiveRecord::Migration[5.1]
-  def change
-    add_column :fields, :choices, :text
-  end
-end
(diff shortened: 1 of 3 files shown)


mboneil10:
https://github.com/umts/stop-project/blob/rewrite/db/migrate/20180614232652_create_fields.rb

mboneil10:
uhhhh

jforgue:
So we don't want to drop the `fields` table?

mboneil10:
```ActiveRecord::HasManyThroughOrderError in BusStops#manage
Showing /Users/molly/stop-project/app/views/bus_stops/manage.haml where line #16 raised:

Cannot have a has_many :through association 'BusStop#routes' which goes through 'BusStop#bus_stops_routes' before the through association is defined.
```

mboneil10:
@jforgue don't think so

jforgue:
If I understand correctly, somehow the `fields` table didn't exist before theses migrations?

mboneil10:
Yes, the `fields` table didn't exist before the two deleted migrations.

mboneil10:
I'm going to look into this association problem before this gets merged.

mboneil10:
Why would this association problem happen NOW. wtf

mboneil10:
Closing for now so no one accidentally merges. Need to figure out the association problem

PR #135, opened 2018-08-29
Better seed data

jforgue:


db/seeds.rb
```diff
@@ -40,33 +43,25 @@
...
+stops.each do |route_number, stops_and_directions|
+  stops_and_directions.each do |direction, stop_names|
+    sequence = 0
     stop_names.each do |stop_name|
-      if direction == 'North'
-        sequence += 1
-      else
-        sequence -= 1
-      end
+      sequence += 1
```
dfaulken: Let's use `with_index` here rather than having a manual counter variable.

db/seeds.rb
```diff
@@ -2,14 +2,14 @@
 
 require 'timecop'
 
-User.create! name: 'David Faulkenberry',
-             email: 'dave@example.com',
+User.create! name: 'Admin User',
```
mboneil10: DAAAAAAAAAMN

db/seeds.rb
```diff
@@ -2,14 +2,14 @@
...
+             email: 'admin@example.com',
              password: 'password',
              password_confirmation: 'password',
              admin: true
 
-User.create! name: 'Jake Foreman',
-             email: 'jake@example.com',
+User.create! name: 'Non-Admin User',
```
mboneil10: okay, yeah. That makes sense.

PR #139, opened 2018-09-05
Fix broken migrations

mboneil10:
For the **rewrite** branch

### DIFF
--- a/app/models/bus_stop.rb
+++ b/app/models/bus_stop.rb
@@ -6,7 +6,6 @@ include DateAndTimeMethods
 class BusStop < ApplicationRecord
   has_paper_trail
 
-  has_and_belongs_to_many :routes
   has_many :bus_stop_fields, dependent: :destroy
   has_many :fields, through: :bus_stop_fields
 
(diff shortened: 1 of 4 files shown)


mboneil10:
The original error I was getting was cannot have a has_many :through association BusStop#routes which goes through BusStop#bus_stops_routes before through association is defined. Or something like that. Paraphrasing.

app/models/bus_stop.rb
```diff
@@ -6,7 +6,6 @@
 class BusStop < ApplicationRecord
   has_paper_trail
 
-  has_and_belongs_to_many :routes
```
dfaulken: But a bus stop does have and belong to many routes, though. Are we saying it's impossible to define this association now that we have a join model?

mboneil10: Idk. We have this line https://github.com/umts/stop-project/pull/139/files#diff-a4c987345e1391c4e1f6ff9be1c6ed5aL18

dfaulken: Oh, I see, sorry. That's fine then. We certainly can't have both.

Issue #138, opened 2018-09-01
Export Data to Hastus


mboneil10:
Either an XML or OIG

mboneil10:
Talked with @dfaulken about how to achieve this a week ago. @sherson and I are waiting to hear back from Laurel.

sherson:
Who has thumbs and forgot to hit send when emailing Laurel? This guy.

I emailed her today.

dfaulken:
Question: what are we waiting on Laurel for?

dfaulken:
@sherson any updates here? I can develop a batch XML export as soon as we know what it needs to look like.

sherson:
No, but I'll reach out to PVTA.

sherson:
PVTA got stuck when they hit the limit of the number of custom attributes in Hastus. They recently got past that hurdle and imported using the CSV (though they want to make more attributes for the purposes of validation). 

I've reached out to them to clarify what they want for a workflow.

werebus:
As @sherson mentioned in the above comment, it seems as if the CSV export is sufficient.

PR #147, opened 2019-04-04
Basic XML export

dfaulken:
Addresses #138.

This way the basic plumbing is there - once we work out all the specs with Laurel, we can complete this.

### DIFF
--- a/app/models/bus_stop.rb
+++ b/app/models/bus_stop.rb
@@ -239,6 +239,44 @@ class BusStop < ApplicationRecord
     end
   end
 
+  def self.to_hastus_xml
+    Nokogiri::XML::Builder.new do |doc|
+      doc.data do
+        all.each do |stop|
+          doc.stop do
+            doc.stp_identifier                    stop.hastus_id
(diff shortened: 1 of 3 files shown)


app/models/bus_stop.rb
```diff
@@ -239,6 +239,44 @@ def self.to_csv(limited_attributes: false)
...
+            doc.stp_ud_sidewalk_less_than_36in    stop.sidewalk_width == "Less than 36'"
+            doc.stp_ud_sidewalk_no_sidewalk       stop.sidewalk_width == 'None'
+            doc.stp_ud_system_map                 stop.system_map_exists?
+            doc.stp_ud_trash_pvta                 stop.trash == 'PVTA'
+            doc.stp_ud_trash_municipal            stop.trash == 'Municipal'
+            doc.stp_ud_trash_other                stop.trash == 'Other'
+          end
+        end
+      end
+    end.to_xml
```
mboneil10: All of these look good to me.

dfaulken:
Thanks for the reviews. I guess I'm open to suggestions on whether we should merge now and adapt later, or just open until we work the full specs out with Jonathan.

PR #148, opened 2019-04-11
Modified factories to use blocks

werebus:
Rather than static attributes (as part of an upgrade). Doing so exposed that someone had modified one of the models without changing the factory.

Also found a method that someone had written, tested, and renamed all in the same PR

### DIFF
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 92.76
+    "covered_percent": 88.24
   }
 }
--- a/spec/factories/bus_stops.rb
+++ b/spec/factories/bus_stops.rb
@@ -2,49 +2,49 @@
 
 FactoryBot.define do
   factory :bus_stop do
-    name 'test_stop'
+    name { 'test_stop' }
     sequence :hastus_id
     trait :pending do
-      name 'Stop Name'
-      bench 'PVTA bench'
-      curb_cut 'Within 20 feet'
-      lighting '20 - 50 feet'
-      mounting 'Structure'
-      mounting_direction 'Towards street'
-      schedule_holder 'None'
-      shelter 'Building'
-      sidewalk_width 'Less than 36 inches'
-      trash 'Municipal'
-      mounting_clearance '60-83 inches'
-      sign_type 'Rectangle (<2014)'
-      shelter_condition 'Poor'
-      shelter_pad_condition 'Great'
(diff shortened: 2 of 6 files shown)


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 92.76
+    "covered_percent": 88.24
```
werebus: As far as I can tell, this is accurate. It seems like it actually may have been a little while since anyone ran the specs though. So...

spec/factories/bus_stops.rb
```diff
@@ -2,49 +2,49 @@
...
+      shelter { 'Building' }
+      sidewalk_width { 'Less than 36 inches' }
+      trash { 'Municipal' }
+      mounting_clearance { '60-83 inches' }
+      sign_type { 'Rectangle (<2014)' }
+      shelter_condition { 'Poor' }
+      shelter_pad_condition { 'Great' }
+      shelter_pad_material { 'Asphalt' }
+      shelter_type { 'Modern half' }
+      shared_sign_post_frta { false }
```
werebus: Changed in #87

spec/factories/bus_stops.rb
```diff
@@ -2,49 +2,49 @@
...
+      bike_rack { 'Bike locker' }
+      real_time_information { 'Yes - Solar' }
+      created_at { Date.yesterday }
+      updated_at { Date.today }
+      bolt_on_base { true }
+      bus_pull_out_exists { true }
+      has_power { true }
+      solar_lighting { true }
+      system_map_exists { true }
+      shelter_ada_compliant { true }
```
werebus: Also in #87

spec/models/bus_stops_route_spec.rb
```diff
@@ -35,12 +35,12 @@
       end
     end
   end
-  describe 'import' do
+  describe 'establish_sequences' do
```
werebus: This was changed (also written) in #111. The name change was suggested in the review process, and then the specs didn't get updated.

werebus: This was changed (also written) in #111. The name change was suggested in the review process, and then the specs didn't get updated.

werebus: This was changed (also written) in #111. The name change was suggested in the review process, and then the specs didn't get updated.

PR #153, opened 2019-07-12
Fix name search

dfaulken:
I didn't write a test, but I did test it. Fixes the issue.

### DIFF
--- a/app/models/bus_stop.rb
+++ b/app/models/bus_stop.rb
@@ -251,7 +251,8 @@ class BusStop < ApplicationRecord
   end
 
   def self.strip_id_from_name(name_with_id)
-    name_with_id[/[^(]+/].strip
+    id_string = name_with_id.match(/ \(\d+\)$/)[0]
+    name_with_id.gsub id_string, ''
   end
 
   private
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ group :development do
   gem 'capistrano-bundler', require: false
   gem 'capistrano-rails', require: false
   gem 'capistrano-passenger', require: false
+  gem 'listen', '~> 3.1'
 end
 
 group :development, :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -133,6 +133,10 @@ GEM
     jquery-ui-rails (6.0.1)
       railties (>= 3.2.16)
     json (2.2.0)
+    listen (3.1.5)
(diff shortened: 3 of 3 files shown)


app/models/bus_stop.rb
```diff
@@ -251,7 +251,8 @@ def name_with_id
   end
 
   def self.strip_id_from_name(name_with_id)
-    name_with_id[/[^(]+/].strip
+    id_string = name_with_id.match(/ \(\d+\)$/)[0]
```
werebus: Why the intermediate step?

```
name_with_id.gsub(/ \(\d+\)$/ , '')
```

werebus:
Also, I know this is pretty far out of the scope, but why are we submitting the "name and id" if we only want the name?

app/models/bus_stop.rb
mboneil10: I'm interested to know too. This was a creative way to do it.

mboneil10:
> Also, I know this is pretty far out of the scope, but why are we submitting the "name and id" if we only want the name?

@werebus, I believe it is for the sake of searching for [stops with the same name](https://github.com/umts/stop-project/issues/122). Not sure if that answers your question.

werebus:
Well, it explains some of the reasoning. But both the way the code was _before_ this PR and _after_ don't actually deal with that possibility.  The search strips off the ID and returns the **a** stop with that name:

https://github.com/umts/stop-project/blob/61aa7bb2eefb6768d77da3556503ebc6db93caee/app/controllers/bus_stops_controller.rb#L56-L60

app/models/bus_stop.rb
dfaulken: I wrote it that way while trying to decode what the original was doing - for the sake of communicativeness I left it with the intermediate variable.

I can condense it if there's a strong preference.

dfaulken:
Good point @werebus. Seems as if we should do the opposite - take the ID out of the string and find by that.

While we're at it, perhaps see if there's a way for the autocomplete to just submit the ID by itself? Then we can remove this controller action and model logic entirely, in favor of the standard `show` routing.

werebus:
If the autocomplete controller method returned an array of objects instead of just an array of values, then you can access the selected object in the [select](https://api.jqueryui.com/autocomplete/#event-select) event handler. That handler could set the value of a hidden field that is also in the form, perhaps?

[Possible inspiration here](https://jqueryui.com/autocomplete/#custom-data).

PR #167, opened 2020-05-29
Name search clean

### LINKED ISSUE #152: Can't name-search for stops with parentheses in the name
I don't remember how to understand what `BusStop.strip_id_from_name` is doing, but it's doing it wrong.

```ruby
BusStop.strip_id_from_name "Old Belchertown Rd (Out) (148)"
#=> "Old Belchertown Rd"
```

BenMelz:
Closes #152
Searching by name now prioritizes an ID if given, then defaults to searching by name if that fails. Also, now the JSON passed to the autocomplete options includes hastus_ids and when an autocomplete option is selected, it automatically redirects to the appropriate edit page.


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 88.24
```
Anbranin, line 3: How did coverage go down 28 percent?

BenMelz: I figured it out - I didn't have chrome installed so a bunch of view testing was getting ignored.

app/models/bus_stop.rb
```diff
@@ -250,8 +250,16 @@ def name_with_id
...
+  def self.find_by_name_search(search_query)
+    id_string_matches = search_query.match(/ +\(\d+\)$/)
+    if id_string_matches.present?
+      extracted_id_string = id_string_matches[0]
+      extracted_hastus_id = extracted_id_string[/\d+/].to_i
+      stop_by_id = BusStop.find_by(hastus_id: extracted_hastus_id)
+      return stop_by_id if stop_by_id.present?
+      search_query = search_query.gsub extracted_id_string, ''
+    end
+    BusStop.find_by(name: search_query)
```
Anbranin, line 262: Finished code golfing:
```suggestion
      hastus_string = search_query.match(/ +\(\d+\)$/).to_s 
      hastus_id = hastus_string.match(/\d+/).to_s
      name = search_query.gsub hastus_string, ''
      BusStop.find_by(hastus_id: hastus_id) || BusStop.find_by(name: name)
```
One thing I should mention is that you don't have to turn strings into integers to search by IDs. Rails is nice and knows what you mean (we do it in the params all the time if you recall; they're always strings).
Like, `User.find "6"` will work just as well as `User.find 6`

coverage/.last_run.json
Anbranin: Oops! The readme _should_ have all the dependencies listed but it's very sparse.

spec/system/index_spec.rb
```diff
@@ -74,9 +74,6 @@
       # in order to be completed by autocomplete
       fill_in 'Enter stop name', with: bus_stop.name.chars.first
       find('div', text: bus_stop.name).click
-      within 'form', text: 'Enter stop name' do
```
Anbranin, line 77: Don't you still need to click the search button for this test to pass?

BenMelz: After implementing the final behavior "the simpler option would be for the autocomplete to just redirect to the show routing using the ID in the autocomplete search string, if possible", it now just redirects when an option is clicked.

Anbranin: I was playing with an older version. 😅 
So now the search button is unnecessary??

app/controllers/bus_stops_controller.rb
```diff
@@ -4,9 +4,11 @@ class BusStopsController < ApplicationController
   before_action :set_fields_for_stop, only: %i[update edit]
 
   def autocomplete
-    stops = BusStop.where 'lower(name) like ?',
-                          "%#{params.require(:term)}%"
-    render json: stops.map(&:name_with_id).sort
+    stops = BusStop.where 'lower(name) like ?', "%#{params.require(:term)}%"
+    jsondata = stops.map { |stop| {label: stop.name_with_id,
```
Anbranin, line 8: use `do/end` in place of `{}` for multi-line blocks (run rubocop to see what I mean)

spec/system/index_spec.rb
Anbranin: Axe it, but on two conditions: Pressing enter should still work (for no-mouse users), and make sure to add some text so the user knows that they have to select a stop to go to it.
I'm not sure removing the search is good from a UI perspective, we'll have to see.

BenMelz: Revisiting this, on second look I don't think we should remove the button. A user could theoretically type something in, then click elsewhere on the page to dismiss the autocomplete. At this point if we remove the button and the controller logic, they can't do anything. If we just remove the button, they still will probably have no idea what to do and even if they figure out that they should re-click on the field and press enter, they would still invoking the same controller logic that the button was previously linked to since the hastus ids is only available from within the autocomplete.

Anbranin: Ahh good thinking. Agreed.

PR #176, opened 2020-12-24
Remove DateTimeMethods lib

werebus:
A lib full of time-formatting methods is something I'd rather not borrow from round-three. Use the Rails custom formatters instead.

### DIFF
--- a/config/initializers/custom_time_formats.rb
+++ b/config/initializers/custom_time_formats.rb
@@ -1,0 +1,5 @@
+# Wednesday, February 5, 2020 - 2:22 pm
+Time::DATE_FORMATS[:long_with_time] = '%A, %B %-d, %Y - %-l:%M %P'
+Date::DATE_FORMATS[:long_with_time] = '%A, %B %-d, %Y - %-l:%M %P'
+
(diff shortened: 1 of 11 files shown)


config/initializers/custom_time_formats.rb
```diff
@@ -0,0 +1,5 @@
+# Wednesday, February 5, 2020 - 2:22 pm
+Time::DATE_FORMATS[:long_with_time] = '%A, %B %-d, %Y - %-l:%M %P'
+Date::DATE_FORMATS[:long_with_time] = '%A, %B %-d, %Y - %-l:%M %P'
+
+Time::DATE_FORMATS[:db_hm] = '%F %R'
```
Anbranin, line 5: I cannot stop thinking of "db" as "database", so what is it really, here?

Anbranin (approved):
This is great

config/initializers/custom_time_formats.rb
werebus: It is "database". That is, this is the same as the `:db` format (Rails built-in), but with no seconds.

PR #180, opened 2021-02-17
Headless tests

werebus:
Setup system tests to use rack-test, headless chrome if JS is needed.

This unearthed some actual test failures, too.

### DIFF
--- a/app/views/users/index.haml
+++ b/app/views/users/index.haml
@@ -8,14 +8,15 @@
       %th Admin?
       %th
       %th
-  - @users.each do |user|
-    %tr
-      %td= user.name
-      %td= user.email
-      %td= check_image user.admin?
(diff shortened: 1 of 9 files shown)


app/views/users/index.haml
```diff
@@ -8,14 +8,15 @@
...
+      %tr
+        %td= user.name
+        %td= user.email
+        %td= check_image user.admin?
+        %td
+          = link_to edit_user_path(user) do
+            %button.btn.btn-default Edit
+        %td
+          = link_to user_path(user), method: :delete do
+            %button.btn.btn-default Delete
```
Anbranin, line 22: Or while we're here, we could change this link to a button...

Anbranin (approved):

PR #242, opened 2023-07-10
Fix "inclusion in" validations

werebus:
`SUPER_HASH` is, unsurprisingly, a Hash. If you iterate through it with only one iteration variable, then you get arrays with two values.

Because of the conditional tacked onto `validates`, this caused these validations to never run.


werebus:
Ugh. There's so much invalid data.

I'll need to address that in this PR.

PR #247, opened 2023-10-03
Remove datepicker

werebus:
There's just one date field in the app, just use a date field.


benmelz (approved):

benmelz (changes requested):
Actually, travis failure looks real

benmelz (approved):

PR #252, opened 2023-10-17
Replace table-based form layouts

werebus:
With Bootstrap (3) compatible ones. They... honestly look a little bad, because of the giant font-size messing up alignment. But, this is an incremental step.


benmelz (approved):
Yeah those giant labels are _rough_ but seems fine for an interim step

PR #255, opened 2023-10-17
Remove `display_name` helper

werebus:
It does way too many jobs, but is only used for one of those jobs.


app/views/bus_stops/edit.html.haml
```diff
@@ -1,3 +1,3 @@
-%h1.page-title Editing #{ display_name @stop }
+%h1.page-title Editing #{ @stop.name }
```
werebus, line 1: We're here in the edit view, so it's definitely a persisted stop, and `name` is validated to be present (and it's not actually user-editable).

benmelz (approved):

PR #265, opened 2023-10-26
Upgrade to Bootstrap 5

werebus:


benmelz (approved):
I'm continuing with my 🤷🏼 approach to the actual styling, but I'll sidenote that with bs5 all labels should have the `.form-label` class (but in reality all that does is set a bottom margin).

werebus:
Yeah, missed that in the upgrade guide.

benmelz (approved):

PR #266, opened 2023-10-27
Add Bootstrap JS

werebus:
And enable navbar collapsing and dismiss-able alerts

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,3 +1,4 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "controllers"
 import "@hotwired/turbo-rails"
+import "bootstrap"
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -9,4 +9,4 @@ Rails.application.config.assets.version = "1.0"
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-# Rails.application.config.assets.precompile += %w( admin.js admin.css )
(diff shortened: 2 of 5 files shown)


app/javascript/application.js
```diff
@@ -1,3 +1,4 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "controllers"
 import "@hotwired/turbo-rails"
+import "bootstrap"
```
werebus, line 4: Bootstrap _does_ have an es-module compatible distribution, too. But, not distributed with the Gem. Maybe it would be better to disconnect the two? It would be the more "modern" way to do JS, at the expense of having to keep the two different Bootstrap packages in sync.

config/initializers/assets.rb
```diff
@@ -9,4 +9,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-# Rails.application.config.assets.precompile += %w( admin.js admin.css )
+Rails.application.config.assets.precompile += %w[bootstrap.min.js]
```
werebus, line 12: This is where the error message told me to put this. It could also go in the `manifest.js` instead. Not sure which is better.

werebus (commented):
A few things here where I'm not 100% sure I'm doing it the "right way"...

app/javascript/application.js
benmelz: Same-source seems much more important to me. If we did want to get the ESM version (which, in my experience, only affects how you import it), I'd vote for either vendoring the whole thing (css included), or `npm`ing it.

benmelz (approved):

PR #268, opened 2023-11-09
Replace other with no sign

### LINKED ISSUE #133: Minor inconsistencies with specifications
Last inconsistencies:
- [x] `need work` has to be `needs work`.
- [x] `date stop checked by` is above stop info instead of below (as well as the time).
- [ ] `other` is a choice for `sign_type`, but it should be `no sign`.

dirschn:
closes #133

### DIFF
--- a/app/models/bus_stop.rb
+++ b/app/models/bus_stop.rb
@@ -44,7 +44,7 @@ class BusStop < ApplicationRecord
     sign_type: ['Axehead',
                 'Rectangle',
                 'MGM and Axehead',
-                'Other',
+                'No sign',
                 'Non PVTA'],
     mounting: ['PVTA pole',
                'City pole',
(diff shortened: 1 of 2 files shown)


werebus (changes requested):
Existing Data needs to be addressed in... some way.

Note that we _tried_ to [validate](https://github.com/umts/stop-project/blob/c721d5ebc2f9f220f13c819257cadff2bad33175/app/models/bus_stop.rb#L175-L179) that the choice is in the list. Of course, we didn't write that validation correctly so it never works...

dirschn:
I had trouble connecting to peek the state of production data, the VPN doesn't seem to like working properly with my cell hotspot. But a janky quick local test suggests this should work assuming all the prod sign types are 'Other'.

I thought about fixing the validation, but I think that should be another PR and I dunno if I actually want to do that 😋

dirschn:
I assume we'd want to include `nil` and `""` in the task as well then

werebus:
No, blank values are allowed (for "incomplete" stops).

werebus (approved):

PR #306, opened 2024-11-01
Add rubocop-capybara

werebus:
And make requested fixes


.rubocop.yml
```diff
@@ -13,6 +14,9 @@ AllCops:
     - 'config.ru'
   NewCops: enable
 
+Capybara/ClickLinkOrButtonStyle:
```
werebus, line 17: This cop is being replaced with `Capybara/AmbiguousClick` at some point. The replacement behaves like `strict`, and I don't feel _that_ strongly one way or the other.

benmelz: Is there justification for adding this? From a writing tests standpoint I do prefer `click_on`

werebus: Yeah, the "justification" is only what I wrote above. rubocop-capistrano is deprecating this cop and replacing it with one that can only act like "strict". All they say is:

> In projects where accessibility needs to be considered, it is crucial to specify the click target precisely.

Which, I dunno, feels kinda right but I can't really say why? To me, it seems (perhaps weakly) connected to, "links are for going places, buttons are for doing things." My thinking here was to move in the direction of disagreeing with the linter less in the future (even if I'm disagreeing more with it in the preset).

All that said. I don't feel that strongly about it, and `click_on` reads better, IMO. I can also change it back.

benmelz: Just to be clear, is it going to become the new _default_ enforced style? If so, then I agree with doing this for the sake of not disagreeing with the defaults more than we have to.

That being said, I can't find any indication that'll be the case - [this thread](https://github.com/rubocop/rubocop-capybara/issues/81#issuecomment-1879035022) (softly) suggests to me that the maintainers don't like the cop at all and added an opt-in replacement rule because someone specifically requested it (in which case I think I'd prefer to not opt-in).

werebus: Yeah, I think I misread [the docs](https://docs.rubocop.org/rubocop-capybara/cops_capybara.html) slightly, what it actually says is:

* This cop is going away
* **If** you were previously using this cop in `strict` mode, then you can enable `Capybara/AmbiguousClick` in the future, but that cop won't be enabled by default.
* If you were using this cop in `link_or_button` mode, then there will be no replacement for that.

---

All that said, I believe there are three options:

1. Start out preferring `click_on` now, transition to no preference in the future
2. Switch to no preference now (disable this cop)
3. Switch to preferring `click_link`/`click_button` now, transition to the new, optional cop in the future.

werebus: After thinking about it for a bit, let's just go with option 1 for now.

If, later, I'm convinced that `click_link`/`click_button` is truly better, then it's not that big of a change to make.

benmelz (approved):

PR #343, opened 2024-11-12
Pull out stop attribute-values hashes

werebus:
Two changes to address some rubocop length/complexity complaints:

1. Moved all of the hashes of field options into a module. Ordinarily, I don't hold with fixing "the class is too long" by plopping stuff into included modules, but in this case, they _are_ clutter, _are_ thematically related, and don't actually have to be `include`d. It's just a file full of `Hash` constants.
2. Refactored `BusStop.to_csv` a bit to decrease length/complexity


app/models/bus_stop.rb
```diff
@@ -210,20 +91,9 @@ def self.to_csv(limited_attributes: false)
...
-                                              'hastus_id',
-                                              'id',
-                                              'updated_at',
-                                              'created_at',
-                                              'route_list',
-                                              'completed_at',
-                                              'completed_by',
-                                              'completed')
-              LIMITED_ATTRS.merge hashed_columns
+              LIMITED_ATTRS.merge columns.to_h { |c| [c.name.to_sym, c.name.humanize] }.except(LIMITED_ATTRS.keys)
```
werebus, line 94: The `except` shouldn't actually matter here except that I'm pretty sure the order of the hash keys matters here. Sure, that's not the greatest idea, but...

app/models/bus_stop.rb
```diff
@@ -210,20 +91,9 @@ def self.to_csv(limited_attributes: false)
...
             end
-    CSV.generate headers: true do |csv|
-      csv << attrs.values
+    CSV.generate headers: attrs.values, write_headers: true do |csv|
```
werebus, line 96: `headers:` isn't a boolean option, it's supposed to take an array of the headers.

benmelz (approved):

PR #656, opened 2026-07-06
Add health endpoint

### LINKED ISSUE #655: Enable health check endpoint
Should be a simple oneliner in `config/routes.rb`.

Also will require a shibboleth public route rule on deploy.

Aryanvkl:
Closes #655.


benmelz (approved):
~~Doing the shib rule now~~

Nope this is another shibless app

